PR #54, opened 2026-07-06
feat(psvita): add PS Vita support

josephinoo:
Hi @doodlewind 
PSVITA SUPPORT
- Adds a native PS Vita runtime (native-vita/) targeting armv7-sony-vita-newlibeabihf: QuickJS host bindings (ffi.rs, exposing globalThis.ui), the main loop (main.rs), pak asset loading (pak.rs), and vita2d-backed rendering (graphics.rs).
- Adds scripts/vita.ts (+ vita npm script) to build the JS bundle/pak and invoke cargo vita build vpk.
- Implements real texture and text rendering on vita2d: graphics::register_texture uploads PSM_4444/PSM_8888 pixel data into GPU textures (SCE_GXM_TEXTURE_FORMAT_U4U4U4U4_ABGR / U8U8U8U8_ABGR) respecting texture stride; graphics::register_font_atlas converts the core's alpha-coverage font atlas into a white RGBA strip texture, and GLYPH_RUN draws each glyph with vita2d_draw_texture_tint_part tinted by the text color. Both pak::feed (boot-time embedded assets) and the ui.uploadTexture/ui.loadFontAtlas FFI ops (runtime JS calls) register into vita2d.
- Fixes a deterministic null-pointer crash (PC=0x0) during JS_Eval: the vendored QuickJS C sources were built with -DJS_NO_NAN_BOXING, but the header never checked that macro, so on 32-bit ARM it silently re-enabled NAN-boxing (JSValue = 8-byte uint64_t) while the Rust bindings assumed the 16-byte boxed struct. Every FFI call returning/taking a JSValue was off by one register under AAPCS, corrupting ctx/rt and eventually branching through a null function pointer inside js_malloc_rt.
- Also raises the Vita main thread stack (sceUserMainThreadStackSize = 1 MB) since the default (256 KB) collided with QuickJS's own default stack budget (JS_DEFAULT_STACK_SIZE) once boot init (vita2d/ctrl/pak) had already consumed part of it.

## Notes 

- The QuickJS header patch itself lives in the vendored libquickjs-sys crate, which was moved out of this repo to ../../quickjs-rs/libquickjs-sys (sibling directory, same layout the PSP target already uses) so it isn't part of this diff. Anyone building the Vita target needs that sibling checkout — happy to also send that patch upstream to quickjs-rs if useful.
- graphics.rs still has no-ops for triangles (TRI) and gradient rects fall back to a solid from-color quad; not part of this PR's scope.

Captura de pantalla 2026-07-06 a la(s) 20 16 07


josephinoo:
@doodlewind

doodlewind:
Hi @josephinoo thanks for your PR! Please give me just a few more days to polish current arch, than I'll scale to more platforms including Vita!

DarkevilPT:
Yoo what happened to the vita ???

PR #93, opened 2026-07-12
feat(style): bevel border rings for classic-chrome themes

doodlewind:
First engine PR of the PocketShell wave: porting the sheru desktop-OS themes (Win98/XP/7/Aqua/XFCE) onto the PocketJS style engine. Win98's signature is the two-ring 3D bevel (98.css inset-shadow stacks) — this adds it as first-class style props.

## What

- **spec**: `PROP.bevelOuterLight/OuterDark/InnerLight/InnerDark` (color) + `PROP.bevelWidth` (f32, default 1) at ids 77–81; ids 73–76 stay reserved for per-corner radius. Not animatable, not layout-dirtying.
- **core**: up to two nested inset rings emitted as plain RECTs in `paint()` — light paints top+left, dark paints bottom+right full-length after it, so **dark owns the shared corners** (matches the 98.css box-shadow stack where the dark shadow is listed first). Square-only: `radius > 0` disables bevels. Consecutive RECTs batch into one sceGuDrawArray on PSP, so a full double-ring bevel is 8 rects ≈ 1 draw call.
- **compiler**: `bevel-[#light,#dark]` (outer ring), `bevel-[#a,#b,#c,#d]` (both rings, order outerLight,outerDark,innerLight,innerDark), `bevel-w-[N]` (arbitrary-only). `bevel-*` + `rounded*` in one literal is a hard error; malformed forms reject the literal (all-or-nothing contract). Pressed inversion is just `active:bevel-[…]` with swapped colors — renders once hosts wire `set_active` (next PR).
- **demo/golden**: `demos/chrome` — a Win98 window mock (double-ring raised frame + buttons, double-ring sunken text well, thin single-ring status cells, navy 2-stop caption gradient) with a d-pad focus step.

## Compatibility

Zero DrawList/backend changes (`ge.rs`/`wgpu`/`raster.rs` untouched); unused props emit nothing. No styles.bin version bump — new props ride the existing flat prop records; old decoders ignore unknown ids.

## Verification

- `bun run test` green (incl. new bevel parser cases + bevel record in the contract round-trip)
- `bun test/golden.ts`: **all 35 existing goldens byte-identical**; 2 new `chrome-main` goldens
- `cargo check`/`cargo test` (core), `bunx tsc --noEmit` clean
- PSP hardware smoke (`bun run hw chrome`) pending — will run together with the setActive PR since that needs an EBOOT rebuild anyway


doodlewind:
PPSSPP e2e pass (run on the stacked #94 branch, which includes this PR): full suite byte-exact incl. 3 new `chrome` shots exercising every bevel form on the native GE path. Real-PSP hardware smoke still pending a PSPLINK cable session.

doodlewind:
Hardware pass done (see #96 for the full notes): bevel rings verified on a real PSP — raised window frame + buttons, sunken text well, thin status cells, and the `active:bevel-[…]` press inversion driven remotely via an injected input tape.

PR #94, opened 2026-07-12
feat(input): wire active: pressed state end-to-end

doodlewind:
Second engine PR of the PocketShell wave (stacked on #93 — retarget to main after it merges). The `active:` style variant has been compiled and resolved since v1, but no host ever triggered it — `hero`'s `active:bg-blue-700` was dead code. This wires it end-to-end so pressed visuals (like #93's Win98 bevel inversion) render natively.

## What

- **spec**: `OP.setActive = 26` — `setActive(id, activeInt)`, same variant machinery as `setFocus`.
- **hosts**: wasm export `ui_set_active` + `wasm-ops.js` binding; QuickJS FFI `js_set_active` on PSP; `op!("setActive", …)` on the wgpu surface. Optional in `HostOps` — older hosts degrade gracefully.
- **input manager** (`src/input.ts`): CIRCLE down holds `active:` on the focused node; release clears (new release-edge detection — `handleFrame` previously early-returned on `pressed === 0`); any focus change while held (d-pad move, FocusScope push/pop, removal repair) clears it so the pressed look never sticks. Zero per-frame JS while held.
- **core fix**: `set_active` on a record **without** an `active:` variant now flips the flag without retargeting — press/release can no longer restart in-flight transitions (CSS `:active` semantics). Without this, the settings golden drifted because a press stretched a mid-flight focus tween.

## Verification

- `bun run test` green, incl. a new renderer test: hold applies `setActive(id,1)`, held frames don't repeat, release clears, d-pad-while-held clears off the old node, CIRCLE with nothing focused emits nothing
- `bun test/golden.ts`: all existing goldens **byte-identical** (settings pins the no-active-variant fix); chrome golden grows f18 (pressed bevel inversion on CANCEL) and f26 — which is **byte-identical to f12**, pinning a clean press/release cycle
- `cargo check` core + pocket-ui-wgpu, `bunx tsc --noEmit` clean
- PSP hardware / PPSSPP pass tracked with #93 (needs the EBOOT rebuild this PR motivates)


doodlewind:
PPSSPP e2e pass complete: full suite **byte-exact** on PPSSPPHeadless (software renderer) — the 7 existing demos are unchanged under the new core+FFI, and 3 new `chrome` shots pin the native path (focused face tint / held-press bevel-ring inversion / release restore). Real-PSP `bun run hw chrome` still pending a PSPLINK cable session.

Note: the PSP EBOOT build was broken independently of this PR by today's local `quickjs-rs` `feat/vita-target` commit (patch.c swapped `#define CHAR_BIT 8` for `<limits.h>`, which misses CHAR_BIT under `clang -target mipsel-sony-psp`). Fixed with a guarded fallback committed locally on that same branch (`~/code/dreamcart/quickjs-rs`, commit `ae957fd`, not pushed).

PR #96, opened 2026-07-12
fix(psp): flush the GE texture cache on every texture bind

doodlewind:
Real-hardware bug found by pocket-shell M1's hardware test (pocket-stack/pocket-shell#1): the Win98 caption-control glyphs — three 16×8 PSM8888 textures drawn back to back — rendered **blank on a real PSP** while every emulator path (wasm rasterizer, desktop wgpu, PPSSPP software AND gles, byte-exact e2e) showed them fine.

**Root cause**: `sceGuTexImage` does not invalidate the GE's hardware texture cache. When consecutive binds change only the texture buffer pointer (same dimensions + format), real hardware keeps sampling stale cache lines; emulators refetch on every bind and hide the quirk. Font atlases dodge it because their binds change size/format.

**Fix**: `sceGuTexFlush()` after each `sceGuTexImage` (both `apply_texture` and `apply_font_texture`). A few binds per frame — negligible against the 8 KB cache refill.

**Verification**: full PPSSPP e2e suite byte-identical (the flush is emulator-neutral); hardware re-test of pocket-shell M1 pending (EBOOT rebuilt against this fix).

Stacked on #94 (pocket-shell's submodule pin includes both).


doodlewind:
**Hardware-verified on a real PSP** (PSPLINK remote session: mailbox screenshot + injected input tape):
- pocket-shell M1's three 16×8 caption-control glyphs now render — dark-pixel count in the controls region **exactly matches the wasm golden (179 = 179)**. Before this fix the same EBOOT rendered them blank.
- chrome demo also hardware-verified: raised/sunken/thin bevels, caption gradient, and the `active:` press inversion (CANCEL top edge white→black while ○ held via injected tape).

Session notes: a stale resident `pocketjs` module from an earlier session held ~50 MB of user RAM and made every `ldstart` fail with `0x800200D9` (NO_MEMORY); `pspsh kill` stops but does not free the arena — a PSPLINK `reset` does.

PR #92, opened 2026-07-12
feat(vita): add native PS Vita runtime and golden E2E

doodlewind:
## Summary

- register a truthful Vita target profile on top of #98: 960x544 physical, 480x272 logical, integer-fit, buttons, left analog, and baked text
- prove a PSP-baseline manifest resolves unchanged for Vita; touch remains unregistered until a public API, HostOps delivery, and tests land together
- route `bun pocket build --target vita` and `bun play vita <demo>` through the shared schema, resolver, small checksummed plan, compiler, and typed backend registry
- make `app.output` the exact artifact name and use stable `HostBuildInputs` for stock and custom native hosts
- keep runtime compatibility to target + HostOps ABI; the plan checksum stays build-time only
- validate target, ABI, 480x272 logical viewport, 960x544 physical viewport, and integer-fit presentation in the Vita backend
- complete the reusable QuickJS + vita2d host, physical controls, left-analog delivery, pak/core texture residency, and exact-2x fullscreen capture path
- recycle same-sized Vita GPU allocations after logical texture retirement, separating generation-handle lifetime from physical GXM storage and avoiding Vita3K's live texture-destruction crash
- make PSP and Vita emulator E2E use the real manifest -> plan -> backend path
- keep DrawList as internal render IR rather than advertising it as a public application capability
- add the draft blog post `One App, Two PlayStations: PocketJS on PS Vita`, including architecture, exact-2x mapping, the texture recycler incident, commands, limitations, and test methodology

## Validation

- `bunx tsc --noEmit`
- `bun run test` (31 contract/type checks, 128 browser/runtime tests, 7 sim tests, 4 deep-zoom tests)
- `cargo test --manifest-path core/Cargo.toml --features std` (58/58)
- renderer/texture browser tests (45/45)
- `bun run golden` (35/35 byte-exact WASM goldens)
- `bun run e2e` (23/23 byte-exact PPSSPP shots through manifest builds)
- `bun run e2e:vita` (35/35, 960x544 exact 2x, GXM textures resident, byte-exact)
- Vita release ELF/SELF/VPK build
- Pocket Figma on the updated contract: 4/4 Vita3K journeys, including mip switch and page switch
- OpenStrike on the updated contract: 5/5 Vita goldens with native scene-counter assertions
- `bun scripts/build.ts hero`, `bun scripts/wasm.ts`, and `bun site/build.ts` (7 blog posts); local preview at `http://127.0.0.1:8140/blog/pocketjs-on-ps-vita/`
[...]

PR #98, opened 2026-07-12
feat(platform): add portable build contracts

doodlewind:
## Summary

- add a strict `pocket.json` v2 schema, validator, plain string capability registry, and one-shot target resolver
- serialize a small `ResolvedBuildPlan` containing only compiler/native consumers; `planHash` is a build-time consistency checksum, not a runtime trust primitive
- make `requires` fail resolution and expose `enhances` through `@pocketjs/framework/platform`
- fold binding-safe `hasFeature("literal")` calls to target booleans in both Solid and Vue Vapor, allowing Bun to remove unavailable branches without minification
- include the sorted target feature map in transform-cache keys so PSP/Vita builds cannot reuse each other's specialization
- typecheck only the app entry and reachable imports with the app's normal TypeScript configuration
- add typed backend dispatch plus `bun pocket check`, `compile`, and `build`
- expose only stable `HostBuildInputs` to custom hosts; the full plan, resolver, profiles, and registries stay internal
- validate runtime pairing with target + HostOps ABI
- define capability admission at the registry: ids name stable public behavior; different execution levels or guarantees use different ids
- remove `ui.drawlist` because DrawList is an internal core-to-backend IR, not an application-observable capability
- intentionally omit capability tokens, target ambient modules, parameter DSLs, package/profile fields, and full-plan runtime hashing

## Validation

- `bunx tsc --noEmit`
- `bun run test` (30 focused schema/resolver/app/platform/host checks, 128 browser/runtime tests, 7 sim tests, and 4 deep-zoom tests)
- Solid and Vue Vapor specialization tests prove alias imports fold, dynamic/shadowed calls do not, cache keys are target-sensitive, and dead sentinels disappear with `minify: false`
- `bun pocket check --target psp`
- `bun pocket compile --target psp`
- `bun pocket build --target psp --manifest pocket.json --project-root . --outdir .pocket/psp-build -- --release`
- schema source and committed JSON Schema are byte-exact
- public manifest/platform subpaths resolve through the app TypeScript config

This is the prerequisite for #92 and registers only the proven PSP profile. The stacked Vita PR adds the Vita profile and proves the same portable PSP baseline at 960x544.


doodlewind:
Merged with review hardening (see fec0a58). Deferred notes for the record:

- `compiler/app-check.ts` hard-imports `typescript`, which is a devDependency — a published-package consumer running `bun pocket check` without typescript installed gets a bare module-not-found. Consider a lazy import with a clear message, or a peerDependency.
- `pocket compile` defaults to the project `dist/`, where it can overwrite a legacy `scripts/build.ts` artifact of the same name with a contract-embedded bundle (harmless today with one target; embed-the-wrong-target once a second target lands). A per-plan default outdir or a marker check would close it.
- `scripts/pocket.ts` uses `new URL(...).pathname` (Windows-hostile, `/C:/...`); `fileURLToPath` for the new cross-project CLI.
- Positional `<app>` alongside `--plan` is silently discarded in build.ts/psp.ts; `--extra-chars` is not forwarded in psp.ts plan mode.
- `validate.ts` counts UTF-16 units where JSON Schema counts code points (only plausibly matters for astral chars in `title`).

PR #129, opened 2026-07-19
feat(pocket-widget): flat widget runtime — Pocket Note, a markdown sticky on the desktop

doodlewind:
> Stacked on #128 (`nifty-podium`) — merge that first; this diff then reduces to the flat-form work.

## What this is

**Pocket Note** — a markdown sticky for the desktop, and the second pocket-widget runtime (the first *flat* one): a borderless, transparent-cornered, always-on-top, resizable window that is nothing but a live PocketJS `ui` surface. No scene, no camera — one `render_words_scaled` pass on dirty frames, governed by the #128 demand-rendering governor.

```sh
bun run note                       # your note (~/.pocket-note.md)
bun run note -- --file todo.md --width 380 --height 520
bun run note --proof               # headless acceptance incl. autosave round-trip
```

- **Preview mode**: parsed markdown (headings, bold/em/inline-code/links, nested lists, quotes, fenced code, rules) with measured styled wrap in a centered max-width column; browser-style drag selection (clicks are inert, exactly like a real preview); list ordinals ride along in copies.
- **Edit mode** (eye/pencil segmented toggle): soft-wrap editor with a real caret, click-to-caret, drag + shift-click + shift-arrow selection, ⌘C/⌘X/⌘V through the system clipboard, and a coalescing ⌘Z/⇧⌘Z undo/redo stack. Edits autosave (debounced) to the backing file.
- **IME input**: OS composition with inline underlined preedit and a candidate window docked at the caret. Glyph coverage is solved at **runtime** — the host rasterizes unseen codepoints from a system CJK font into the pak's FONT ATLAS v3 blobs and reloads them through the spec `loadFontAtlas` op. Typing 你好 bakes exactly two glyphs, on the spot.
- The `•••` menu (Portal overlay): preview/edit, theme, reset, close — outside clicks close it anywhere, including over the header drag region.

## The architecture call

The capability stack was already 80% there; this PR fills exactly the missing half-layers, **zero spec changes**:

- **`FlatWidget` + `run_flat`** (pocket-widget): the #128 governor behind one internal `Driver`; `WidgetConfig` gains `resizable`/`min_size`/`ime` + a `resize_at` grip hook and an `ime_cursor_area` hook.
- **A resize is a relayout, never a zoom**: the host renders at the window's true scale factor and live-resizes the core viewport (`Ui::set_viewport` + the new framework `resizeViewport()`); text reflows against the new width with zero font-size jitter.
- **The svc channel (spec ops 30..32) is the desktop companion contract**: keyboard/mouse/wheel/resize/IME in as JSON lines; save/quit/menu/copy/caret intents out. The same bundle still boots on svc-less hosts (PSP, sim, goldens) as a read-only d-pad note — the unmodified-app base case holds.
- **Clicks are CIRCLE**: hover→focus from svc mouse moves + the framework's stock onPress pipeline dispatch all chrome; content gestures (caret, drag-select) ride the mouse stream, and sub-16ms clicks are edge-detected so none drop.
- **Retina text the Vita way**: `render_words_scaled` + density-2 atlases sample 1:1 (`UiSurface::new_with_density`, `build.ts --density=N`).

## Framework fixes surfaced by the widget

- Hit tests now span the **overlay layer**, and scope filtering applies only while a `FocusScope` is pushed — Portal/Modal content was unclickable for every cursor-mode host before this.
- `Focusable`/`Screen`/`ActionBar` froze dynamic class/style at mount (`{...props}` spread reads Solid's compiled getters once) — now `mergeProps`.
- `Portal` sizes to the live `__viewport` instead of hardcoded 480×272.
- `pocket3d::Input` grows edit-keystroke/IME/wheel streams; the wgpu renderer re-uploads font slots whose glyph count moved.

## Power: measured, and the receipt explains itself

```
pocket-widget: 361 ticks, 2 frames rendered (0.6%) — armed by dirt 1, resize 2, occlusion 1, scale 1; 0 unarmed OS redraws skipped
```

Idle steady-state: **~0.7–1.1% CPU, ~86 MB RSS, one process, zero frames after settle.** The receipt names its arm sources (in steady state the only healthy one is dirt), and OS-initiated redraws with nothing pending are skipped — a settled widget can't be turned into a space heater by compositor moods.

## Verification

- `test/note.test.ts` (33): markdown parser, offset-preserving wrap + caret math, selection ops, undo/redo coalescing, preview selection mapping + copy extraction — all pure, fake-measure.
- Umbrella suite green; **49/49 pixel goldens** from clean `dist/`; cargo workspace tests green.
- Headless end-to-end (screenshots + file/clipboard assertions): view render, click-to-caret + typing, drag/shift-click selection both modes, delete→undo→redo round-trips, ⌘C→pbpaste byte-comparison, paste/cut, menu open + outside-click close, IME preedit underline + 你好世界 commit→autosave, full Chinese note rendering, light theme.

## Known limits (documented in WIDGET.md §9)

Bold slots borrow the regular CJK fallback face; no line-start kinsoku for CJK wrap; no mono font slot for code blocks.


doodlewind:
**Round 2** (`d528a01`) — review feedback:

- **Icon mode toggle**: EDIT/DONE text buttons → a segmented eye / I-beam toggle (icons are composed Views; the atlas has no icon glyphs). Gray header dot removed.
- **Preview drag-select**: browser-style blue selection over the rendered rows (`select.ts`, (row, char) space — boundary rows clip to the drag endpoints, code blocks select atomically, rules are inert). A drag never enters edit mode; a clean click (release without drag) still does.
- **Editor drag-select**: per-line highlight rects + a newline sliver on fully-selected empty lines; Backspace/Delete eat the selection, typing replaces it.
- **Undo/redo**: snapshot stack with run coalescing (a typing run = one step, a delete run = one step, caret moves break runs), ⌘Z / ⇧⌘Z. Undone states autosave.
- **Menu outside-click**: the guest reports `{t:"menu", open}` and the host stops claiming header drag/resize while it's up, so any click lands on the backdrop and closes it.
- Protocol: svc mouse lines now carry the primary-button state (press/release edges even without motion); headless harness gains `--drag x0,y0-x1,y1@frame`.

Verified: delete-selection → autosave → Undo → file byte-restored → Redo re-applied (file assertions); drag-select screenshots both modes; menu open + header-click close. 30 note tests, umbrella suite, 49/49 goldens, idle receipt unchanged (2 frames / 361 ticks).

doodlewind:
**Round 3** — preview interaction corrected: content clicks are now inert (a markdown preview doesn't react to clicks; edit mode is entered via the eye/I-beam toggle or the menu). Drag selects, **⌘C copies** in both modes — the guest sends a `{t:"copy", text}` intent and the host pipes it to the system clipboard. Soft-wrapped rows re-join with the wrap-consumed space on copy, so a wrapped paragraph pastes as one paragraph; explicit newlines, block boundaries and code blocks keep their line structure. Verified end-to-end with a pbpaste byte-comparison (user clipboard saved/restored around the check). 32 note tests, 49/49 goldens, idle receipt unchanged.

doodlewind:
**Round 4** — resize/layout/clipboard polish: (1) **grip-resize jitter fixed** — the host was rendering at a window-px/viewport ratio, which disagrees by sub-pixel rounding mid-resize and re-scaled every glyph (~0.3% shimmer); rendering now always uses the window's true scale factor, so a resize is purely a relayout, never a zoom. (2) **Centered content column** — caps at 560 logical px and centers in wider windows, base side padding 14→22 px. (3) **⌘V / ⌘X** complete the clipboard: paste reads pbpaste host-side and inserts as one undoable step (replacing any selection); cut = copy + delete. Verified with file/clipboard assertions and a 720px-wide screenshot; 32 note tests, 49/49 goldens, idle receipt unchanged.

doodlewind:
**Round 5** — (1) **Real icons**: outlined eye + 45° pencil as baked SVGs (bakeSvg's filled-path + hole-punch subset; per-theme ink, active segment keeps the bg) replacing the composed-rect icons that read as a switch. (2) **Flaky selection-not-cleared click fixed**: a sub-16ms click lands both edges inside one 60 Hz tick and level sampling dropped it entirely; the host now reads the press edge and emits both svc mouse edges + a one-tick CIRCLE, so every click reaches the guest. (3) **List markers in preview copy**: ordinals/bullets ride along when the selection covers the line start (`1. first`). (4) Framework fix surfaced by the toggle: `Focusable`/`Screen`/`ActionBar` spread `{...props}`, freezing dynamic class/style getters at mount — now `mergeProps`. 33 note tests, 49/49 goldens.

doodlewind:
**Round 6 — IME input.** Composition works end to end: the shell enables OS text input (`WidgetConfig::ime`) and streams winit's Preedit/Commit through the Input; the guest splices the preedit at the caret with an accent underline and reports its caret rect back so the candidate window docks next to the text (`set_ime_cursor_area`). On macOS plain typing stays on the KeyEvent path while composition suppresses it — disjoint by construction, no double input.

The interesting half is glyph coverage: instead of guessing a CJK charset at build time (and shipping megabytes of hanzi), the host **extends the font atlases at runtime** — it mmaps a system CJK font, rasterizes unseen codepoints with ab_glyph into each slot's fixed cell at pak density, appends them to the FONT ATLAS v3 blob (cmap stays sorted; coverage is gid-linear, so appending is cheap) and reloads the slot through the spec `loadFontAtlas` op. The wgpu renderer re-uploads any slot whose glyph count moved. Typing 你好 bakes exactly two glyphs, on the spot, across all seven baked slots.

Verified headless: preedit underline + real glyphs at the caret, a full Chinese note (heading/body/code/list/quote) rendering in preview from the load path, and a commit → Escape → autosave round-trip of 你好世界 to disk. 33 note tests, 49/49 goldens, idle receipt unchanged. v1 limits: bold slots borrow the regular fallback face; no line-start kinsoku for CJK wrap.

doodlewind:
**Round 7** — shift-extend selection: svc mouse presses and named keys now carry the shift modifier. Edit mode: shift-click keeps the anchor and moves the caret; shift+arrows/Home/End extend the same way. Preview: the anchor persists across plain clicks, so click → shift-click selects the span between the two points (and shift+drag keeps extending). Headless `--shift-click x,y@frame`; verified with a cross-block span screenshot and an extend-then-delete file assertion. 33 note tests, 49/49 goldens, receipt unchanged.

doodlewind:
## Measured: CPU + memory footprint

Steady-state methodology (pocket-character's): release build, real window on an M3 Max (Retina 2×, density-2 assets), 75 s runs, `ps` sampled every 2 s with the first 12 s (boot) discarded, pointer away from the window. One process, 6 threads, no children; binary 8.9 MB, pak 1.07 MB.

| Scenario | CPU avg | CPU max | RSS | Governor receipt (ticks → frames) |
| --- | --- | --- | --- | --- |
| **Idle**, preview | **0.76 %** | 1.4 % | **87.1 MB** | 4501 → **2** (0.0 %) — armed by dirt 1 |
| **Editing**, caret pulsing | 2.11 % | 5.9 % | 87.8 MB | 4501 → 1568 (34.8 %) — armed by dirt 3712 |
| **Scrolling**, one step / 0.5 s | 1.58 % | 2.5 % | 89.8 MB | 4501 → **138** (3.1 %) — armed by dirt 137 |
| Idle, **CJK note loaded** | 0.44 % | 0.6 % | 93.3 MB | 2701 → 2 (0.1 %) — armed by dirt 1 |

Reading the receipts:

- **Idle is the design goal working**: 75 seconds on screen, two GPU frames total (boot + hello relayout), sub-1 % CPU — that's the fixed 60 Hz QuickJS guest turn over a settled DrawList, nothing else.
- **Scrolling shows demand rendering exactly 1:1**: 137 scroll steps → 137 dirty frames, nothing rendered between steps.
- **Editing is the current ceiling** and it's the blinking caret: `animate-pulse` runs core-side, so most ticks report dirt and the governor paces it to ~21 fps. 2 % CPU while animating is fine, but a discrete blink (two toggles/s instead of a continuous opacity ramp) would collapse this to ~2 % of frames — noted as an easy follow-up.
- **The runtime CJK path costs +6.2 MB** over baseline (mmapped system-font pages touched by rasterization, 31 glyphs × 7 slots of coverage + re-uploaded atlas textures) and its idle profile is unchanged.

For scale: the pocket-character measurement of the same stack's 3D form was 1 process / 118 MB / 3.9 % CPU *while rendering skinned 3D at 60 fps continuously* — the flat form's numbers sit where you'd hope: a fraction of that at rest, and rest is what a sticky note does all day.

doodlewind:
## Measured: the same widget as a Chromium webview (Electron bench)

Built the fair comparison and ran it through the identical methodology (75 s runs, 2 s `ps` sampling, first 12 s discarded, same machine): a minimal **Electron 41** (Chromium ~132) frameless/transparent/always-on-top 420×560 window, the same sample note rendered by `marked` into a `contenteditable`. Numbers are sums over the whole process tree.

| | Pocket Note (measured) | Electron bench (measured) |
| --- | --- | --- |
| Processes | **1** (6 threads) | **5** (main, renderer, GPU, utility, crashpad) |
| Physical footprint (Activity Monitor) | **34 MB** | **~84 MB** (41 + 18 + 18 + 6.7) |
| `ps` RSS (sum) | 87 MB | **351–363 MB** |
| CPU — idle | 0.76 % | **0.29 %** |
| CPU — caret blinking | 2.11 % | **0.24 %** |
| CPU — step scroll ×2/s | 1.58 % | **0.11 %** |
| Disk (framework + app) | 8.9 MB + 1 MB pak | 295 MB dist + app |

Honest reading, both directions:

- **Memory/processes/disk are structural wins for the Pocket stack**: 2.5× on true footprint, 4× on RSS-sum (both metrics overcount shared pages — RSS counts Chromium's framework pages once *per process*), 5× on process count, ~30× on disk. This is the floor Chromium can't go below; it's the price of process isolation plus V8/Blink generality.
- **Chromium wins every CPU micro-scenario**, and it's worth being precise about why: Blink is fully event-driven — an idle page truly sleeps — while our guest pays the Law 3 determinism tax (a 60 Hz QuickJS turn forever, ≈0.7 %). The caret line is our own bug-of-choice: their caret is a 2 Hz dirty-rect blink, ours is a continuous `animate-pulse` opacity ramp that forces ~21 fps (already noted as a follow-up — a discrete blink closes most of the gap). Step-scroll is compositor layer movement for them (pre-rastered) vs a full 60-op DrawList frame per step for us; ours is still only 1.6 %.
- Where the ledger nets out for a **widget that sits on screen all day**: both idle under 1 % CPU — indistinguishable in energy terms — while the memory gap (34 vs 84 MB real, ×N when you open several notes: each Chromium window adds a renderer) and the 295 MB framework are permanent.

### Where Pocket Note's memory actually goes

`vmmap`/`heap`/`footprint` on the live release process: the oft-quoted **87 MB RSS is ~52 MB shared dyld-cache framework pages** (Metal/AppKit, machine-wide, counted per process by `ps`). The true physical footprint is **34 MB**:

| Region | Size | What it is |
| --- | --- | --- |
| malloc heap (live) | ~12 MB | ~70k nodes; dominated by 16–96 B QuickJS objects (Solid graph + bundle ≈ 5 MB), 1 MB pak copy, ~2 MB CPU-side font-atlas coverage (cjk.rs keeps an appendable copy), DrawList + wgpu structs |
| malloc slack | ~4 MB | dirty-but-free zone pages |
| IOSurface | 7.5 MB | the swapchain: two 840×1120 BGRA buffers, to the byte |
| GPU resident (IOAccelerator) | ~1.5 MB | atlas/image textures |
| framework dirty __DATA / stacks / misc | ~3 MB | |

Boot shows a **440 MB transient** (Metal pipeline compilation + eval arenas) fully released within ~1.5 s — steady state is flat at 34 MB. Obvious future trims: rebuild-on-demand instead of retaining the cjk.rs atlas copy (−2 MB), single-buffer the swapchain if wgpu allows (−3.7 MB).

doodlewind:
**Round 8 — the desktop surface is now first-class in the platform contracts** (review feedback: a PSP can't run this, and the framework should be able to say so).

Six new registry capability ids name what the widget actually provides — `input.text`, `input.pointer` (a real pointer is deliberately NOT `input.cursor`: different guarantee, different id), `input.ime`, `host.clipboard`, `display.viewport.live`, `text.glyphs.runtime` — and a new **`desktop-widget-macos`** target profile (hostAbi 3, density 2, `dynamicViewport` range) provides them. Host-windowed target ids follow **`<class>-<form>-<os>`**, leaving room for `desktop-app-macos` / `desktop-kiosk-linux`.

The note's new `pocket.json` is the reference dual-nature manifest: the desktop surface sits in `enhances`, so the same source **admits everywhere** — on PSP it resolves with those features off and the app builds as a read-only note (no pencil; `hasFeature()` gates are compiler-foldable, so the PSP bundle sheds the editor as dead code). A widget-only app that puts them in `requires` is **refused by PSP admission at resolve time** (`capability.unavailable`), not discovered broken on device. `bun run note` now builds through the manifest plan (the `--density` flag hack is retired — density and features come from the profile), and the widget host asserts its identity (`__host`/`__hostAbi`) against plan-built bundles.

21 contract tests (admission both ways, dynamic-viewport ranges), 33 note tests, 49/49 goldens, `--proof` green. One noted opening: the manifest carries a single viewport, so a real PSP build of the note would want a per-target viewport override — left for when someone actually ships one.

doodlewind:
**Round 9 — truthful per-demo manifests + the site filter.** Every stock demo now carries its own `pocket.json` (requires minimized to actual usage; the borrowed hero template over-required `input.analog.left` for everything — real enhances turn out to be just im's nub scroll and the cursor demo's `input.cursor`; hero-vue-vapor finally declares its framework). `demoManifestFor()` makes committed manifests win over template synthesis in `vita.ts`/`play.ts`. The site's demo shelf applies the admission rule — only psp-resolvable manifests are listed, which is exactly how the desktop-only note stays off the landing page and playground. A table-driven **admission matrix** test pins every demo × every target (console demos: psp ✓ vita ✓ desktop ✗; note: the inverse), and a new demo without a matrix row fails CI on purpose. All suites green.

PR #137, opened 2026-07-20
chore: upgrade Vue to 3.6.0-rc.1

zhiyuanzmj:
1.  Upgrade vue-jsx-vapor to 3.2.19 to support Vue@3.6.0-rc.1.
2. Use a functional component instead of defineVaporComponent so you can access props directly instead of attrs, without needing to define runtime props.
3. Fixed an undefined `node.children` error in devtools.ts for Vue Vapor support.
    image


doodlewind:
Verified both Copilot findings empirically and pushed a follow-up commit (53ce906) addressing them.

**How it was verified**: built a minimal probe app using the exact `callbackProp` pattern and ran it in the deterministic sim host with a scripted button press, counting invocations through globals (the UI tree lags in the sim, so counters are the ground truth):

- `typeof props.onToggle` inside a functional component → `"function"` — the vapor attrs proxy resolves the compiler's prop getters before the component sees them (`instance.props === instance.attrs` for function components in `@vue/runtime-vapor`, and attrs `get` runs `resolveSource`).
- One CIRCLE press → **`onToggle` fired twice**: `callbackProp`'s zero-arg probe invoked the already-resolved handler, then the `onToggle()()` call site invoked it again. For Settings this cancels the parent's `sfx`/`vibration` flip while the knob animates — a real double-fire, exactly as flagged.
- After the fix (direct `props.onToggle()`): one press → one invocation.

The fix drops the now-dead `propValue`/`callbackProp` indirection across the four demos that still carried it (settings, gallery, cards, stats) and calls callbacks exactly once, which also removes the dead `typeof props.frame === "function"` branch in stats. All demo variants rebuild cleanly (solid + vue-vapor), the full test suite passes, and settings/gallery/cards/stats/library vue-vapor bundles run a scripted input journey in the sim without errors.

Thanks for the upgrade — the functional-component form is a real simplification. 🙏

PR #140, opened 2026-07-21
feat: configurable viewport, fused ARGB raster, and Vue Vapor host hardening for ESP32-P4

odex21:
## Motivation

ESP32 devices ship with wildly different panels — 480×272, 640×226, 800×480, etc. — there's no "standard" size like a PSP screen. PocketJS currently hardcodes 480×272 in the rasterizer, the wasm binding, and both hosts. This PR removes that assumption, then adds the last few pieces needed to run on real ESP32 hardware and iterate in the browser preview.

Each commit is self-contained and reviewable independently.

## Commits

1. **`feat(core,wasm,host): configurable viewport for layout and raster`** — `Ui` carries a viewport; `render_scaled` derives framebuffer dimensions from it instead of the fixed `SCREEN_W/H`; wasm exposes `ui_set_viewport`; web/sim hosts accept viewport and raster-density options. Any target can size its logical surface at boot without touching the core. Stock hosts keep the 480×272 golden contract.

2. **`feat(raster): fused ARGB8888 output for ESP32-P4 PPA`** — new `render_scaled_argb()` emits A,R,G,B bytes — the in-memory order the ESP32-P4 PPA SRM consumes as ARGB8888. Byte placement is fused into `blend_px` via a const generic shared with the RGBA path, so it costs nothing; the firmware drops its per-frame RGBA→ARGB CPU reorder copy (measured: present 25.5 → 13.0 ms, 14 → 17.1 FPS).

3. **`fix(vue-vapor): isolate the guest document and tolerate fragment-wrapper mirrors`** — covers two issues encountered while hosting a Vue Vapor player page:
   - **Document isolation**: when PocketJS is embedded in a real browser page, `globalThis.document` already exists. The old `if (!g.document)` guard never fires, so the PocketJS DOM facade is not installed, and the Vue Vapor guest's `document.createElement()` falls through to the real browser DOM. This PR introduces `globalThis.__pocketDocument` and aliases `document` to it via esbuild `define` (only for `--framework=vue-vapor` builds), so the real `document` stays untouched.
   - **Fragment-wrapper tolerance**: Vue Vapor injects internal fragment wrappers (`{ nodes: […] }`) into `node.children` during dynamic branch swaps (e.g. `{playing ? <PauseIcon/> : null}`). Upstream PR #137 and #142 add `if (node.children)` guards that prevent crashes when `children` is undefined — but a fragment wrapper passes the truthy guard and is silently serialized as `{"t":"undefined"}`, losing all inner content. `forEachTreeChild` / `forEachTreeMirror` recursively unpacks these wrappers so tree snapshots stay correct.

## Why `forEachTreeChild` instead of `if (node.children)`?

PR #137 (merged) and #142 (open) add null guards to `serializeNode` / `countNodes` / `subtreeHasRetained`. These prevent the crash but don't handle the fragment-wrapper case. A test added in this PR demonstrates the difference:

```ts
// Inject a Vue Vapor fragment wrapper into the tree
(playbackStatus.children as unknown[]).push({ nodes: [[pausedTextNode]] });
// Trigger tree serialization
push({ t: "getTree" });
frame();
expect(treeJson).toContain("PAUSED"); // ← fails without forEachTreeMirror
```

With the simple guard: wrapper passes through → serialized as `{"t":"undefined"}` → **test fails**.  
With `forEachTreeChild`: wrapper is recursively unpacked → `"PAUSED"` found → **test passes**.

This PR keeps the comment-anchor skip from #137 — `forEachTreeChild` subsumes both.

## Compatibility

- `render` / `render_scaled` output is untouched and byte-exact: **49/49 committed pixel goldens pass**.
- `cargo test` (core): **78/78**, including a new test pinning ARGB↔RGBA channel mapping through the translucent destination-read blend path.
- `bun test`: renderer / clock / tiles / cursor / devtools / vue-vapor-dom **88 pass**; sim **7 pass**; launcher-sim **10 pass**.
- No new dependencies, no feature flags, no breaking API changes.

## Hardware validation (commit 2)

ESP32-P4 @ 640×226 → PPA 2× scale + 90° rotation → 452×1280 panel: present time **25.5 → 13.0 ms**, whole-pipeline **14 → 17.1 FPS**, pixel stream bit-identical to the previous software reorder (colors verified on device).


odex21:
## Updated PR description

The PR body was revised to address review feedback. Key changes:

**Removed:**
- Duplicate rebase notes (the "package.json test chain is a union" and "src/devtools.ts" paragraphs appeared twice)
- The individual "feat(host-web): viewport URL params, pointer touch input, perf HUD" commit was folded into the viewport commit description — its content is still in the diff, just not called out separately since it's integral to the viewport feature

**Added:**
- A dedicated section **"Why `forEachTreeChild` instead of `if (node.children)`?"** that explains the difference from the upstream guard pattern in #137 and #142
- A concrete fragment-wrapper test snippet showing the failure mode with the simple guard
- Clearer separation between the two halves of commit 3 (document isolation vs fragment-wrapper tolerance)

**Unchanged:**
- All three commits, their code, and their commit messages are untouched
- Hardware benchmarks and compatibility numbers are the same

PR #145, opened 2026-07-22
feat(vue-vapor): support Vue single-file components

odex21:
## Why

PocketJS already supports Vue Vapor through JSX, but Vue Vapor JSX is a niche authoring format that coding agents rarely generate correctly. Even a small component requires authors to remember `defineVaporComponent`, runtime prop declarations, `setup()`/slot conventions, and TSX-specific details. The failure modes are concrete: function components without a runtime props declaration receive empty props in production builds and crash far from the cause; the setup idiom `return () => …` breaks vapor function components, which must return the block directly; refs need explicit `.value` (no template auto-unref); React habits leak in (`className`, children-as-props instead of slot functions, half-wired `v-model` missing `onUpdate:modelValue`); and the JSX path has thin docs and weaker type inference than SFC templates. Vue SFC syntax is much more common in training data and therefore gives both humans and agents a reliable way to author Vue components.

This does not add the VDOM renderer. Selecting `framework: "vue-vapor"` makes PocketJS compile every imported `.vue` file directly to Vapor, without requiring a `vapor` attribute in the SFC.

## What changed

- compile `<script setup>` + `<template>` files with `@vue/compiler-sfc`
- collect template classes and text during PocketJS pass 1
- allow compiled SFC components to mount directly through the Vue Vapor renderer
- normalize Vue template attrs at the PocketJS host-component boundary
- teach `pocket check` to accept `.vue` imports via an ambient `*.vue` module declaration
- add an SFC port of the Hero demo and a small feature lab covering `v-model`, `v-if`/`v-else`, `v-for`, props, events, and slots
- document the supported SFC subset and include both demos in build/test coverage

PocketJS continues to reject runtime `<style>` blocks, template preprocessors, external blocks, and Options-API-only components.

## Validation

- `bun run test` — 387 passed, 0 failed
- Hero SFC and Vue SFC feature lab both build for the Web simulator
- headless feature-lab test exercises model updates, conditional branches, keyed lists, props/emits, and slots


odex21:
@codex

doodlewind (approved):
Reviewed and verified locally. The design is right: Vue's own @vue/compiler-sfc (vapor mode, inline templates) owns SFC compilation, so PocketJS pass-1 collection sees static classes/text as plain string literals with no template-specific walk, and the unsupported subset (style blocks, external src, preprocessors, options-API-only) fails loudly at build time.

Two things were fixed while landing:
- Resolved the #149 one-axis-per-directory restructure conflicts: new demos moved to apps/, tests to tests/, compiler/runtime files to framework/, and package.json/README/pocket.json paths updated to the new tree.
- Fixed a latent type error this PR exposed: hero-vue-sfc is the first app whose per-app TypeScript check resolves @pocketjs/framework/vue-vapor directly (JSX apps import @pocketjs/framework and get remapped at build time), which pulled framework/src/prelude.ts into a checked graph for the first time — its setTimeout stub cast failed TS2352 against Bun's Timer typings. Now converts through unknown.

Validation: full bun run test — 387 pass / 0 fail, including the new tests/vue-sfc.test.ts compiler coverage and the vue-sfc-lab sim run (v-model, v-if/v-else chains, keyed v-for, props/emits/slots); pocket compile --target psp passes for both new apps; launcher covers regenerate cleanly; lockfile verified stable.

PR #148, opened 2026-07-22
fix(core): prevent gaps in scaled rounded boxes

odex21:
## Summary

- quantize a flat rounded box's shared outer edges once before splitting it into corner quads and rectangular bands
- clamp the destination corner radius to the quantized dimensions
- add a raster regression for a `scaleX`/`scaleY` 0.96 rounded box

## Root cause

The flat rounded-box fast path emitted four textured corners and three rectangles from fractional transformed coordinates. DrawList stores integer positions and integer sizes, so independently rounding each piece's start and extent could leave adjacent pieces one pixel apart. A pressed button using `scale: 0.96` exposed the gaps as a black cross through its background.

The fix quantizes the outer boundaries once and derives every piece from those same boundaries. The fast path still emits the same seven draw operations and performs no additional allocation.

## Visual comparison

| Before: pressed button in the Gesture page | After: isolated pressed rounded box with the core fix |
| --- | --- |
|rounded-box-scale-before|rounded-box-scale-after|

## Validation

- `cargo test --manifest-path core/Cargo.toml scaled_flat_rounded_box_has_no_gaps_between_fast_path_pieces -- --nocapture`
- `cargo test --manifest-path core/Cargo.toml` (82 passed)
- reproduced in the Web host with Vue SFC and TSX versions of the same scaled rounded button


doodlewind (approved):
Reviewed and verified locally after resolving the #149 restructure conflict (merged main; the fix now lives at engine/core/src/draw.rs).

- Root cause confirmed: xy_word/wh_word round each piece's position and extent independently, so fractional transformed edges let adjacent fast-path pieces disagree by 1px. Quantizing the four outer edges once and deriving all seven pieces from them is the right fix, and the added rf clamp correctly guards the degenerate case where quantization shrinks the box below 2r.
- Verified the regression test fails against the unfixed draw.rs and passes with the fix.
- cargo test (engine/core): 82/82. Full bun run test suite: pass. tape:check: 180/180 hero frames unchanged — integer-coordinate output is byte-identical, as intended.

PR #155, opened 2026-07-23
feat(build): --font-regular/--font-bold source overrides

odex21:
## Summary

`--font-regular=<path>` and `--font-bold=<path>` CLI flags let a target build with alternative TTF sources without replacing the shared Inter fonts in `assets/fonts/`.

## Motivation

`framework/compiler/bake-font.ts` already accepted `regularTtf`/`boldTtf` in `BakeOptions`, but `tools/build.ts` never wired them up to CLI args.

## Use case

The ESP32-P4 lyrics firmware needs a subset CJK font to fit the app partition, while desktop/web builds keep the stock Inter fonts.

## Changes

`tools/build.ts` — 9 lines added:
- `--font-regular=` and `--font-bold=` argument parsing
- Forward to `bakeAtlases({ regularTtf, boldTtf })`

### DIFF
--- a/tools/build.ts
+++ b/tools/build.ts
@@ -17,6 +17,9 @@
 //   --config=<path>              load a Pocket config file (default: pocket.config.ts)
 //   --no-config                  ignore pocket.config.ts
 //   --extra-chars=<string>       force extra codepoints into every atlas
+//   --density=<integer>          low-level target sampling density (default 1)
+//   --font-regular=<path>        override the regular font source
+//   --font-bold=<path>           override the bold font source
 //   --outdir=<path>              write <app>.js/.pak here instead of dist/
 //                                (external repos build their apps against a
 //                                vendored PocketJS and keep outputs local)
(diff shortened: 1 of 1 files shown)


doodlewind (approved):
Reviewed and validated locally on a merge with current main:

- `bun test tests/font-bake.test.ts` passes; explicit `--font-regular/--font-bold=assets/fonts/Inter-*.ttf` produces **byte-identical** output to the default build (default path provably untouched), and a genuinely different TTF changes the baked pak — the option flows through end to end.
- Flag parsing and one-arg `resolvePath` (cwd-relative) match the existing `--plan/--project-root/--outdir` conventions, and `bake-font.ts`'s `?? DEFAULT_*` keeps Inter as the fallback.
- Nit, non-blocking: a bad path surfaces as a raw ENOENT from `loadFont` — a one-line existence check with a friendly message would be nicer.

Thanks for also documenting the pre-existing `--density` flag.

PR #172, opened 2026-07-23
feat(hosts): add PocketBook inkview e-ink host

wonderbeel:
## Summary

A new PocketJS host that renders the UI runtime on **PocketBook e-readers** via the [inkview](https://github.com/simmsb/inkview-rs) SDK, plus the two small shared changes it builds on. Design is documented in `pocketjs-inkview-implementation.md` (companion to the `pocketjs-inkview-integration.md` feasibility study).

## Commits

1. **`refactor(engine): extract backend-agnostic pocket-ui-surface crate`** — moves the renderer-independent `ui` surface (`UiSurface`, pak walker, DevTools mailbox) out of `pocket-ui-wgpu` so non-GPU hosts can reuse it without dragging in wgpu. `pocket-ui-wgpu` re-exports it; desktop consumers are source-compatible.
2. **`feat(pocket-mod): add Guest::frame_with_touches`** — additive 3-arg `frame(buttons, analog, touches)` path (the framework already decodes a packed-touch array; pocket-mod had no way to deliver it).
3. **`feat(hosts): add PocketBook inkview e-ink host`** — the standalone host (same lone-bin model as `hosts/psp`/`hosts/vita`).
4. **`docs: integration + implementation guides`**.

## How the host works

- Reuses `UiSurface` for the entire `ui` HostOps surface + pak feeding, and the core's software rasterizer (`render_scaled`) at `logical × density`.
- `framebuffer.rs`: RGBA8 → Gray8 luminance + 16×16-tile damage.
- `refresh.rs`: e-ink update policy ported from `inkview-slint` (partial when idle, throttled dynamic during motion, 200 ms quiet cleanup).
- `input.rs`: inkview keys → spec BTN bits; touchscreen → packed touch wire format.
- `main.rs`: `iv_main` forwards events into a channel; a second thread owns the `Screen` + fixed-cadence tick/render loop (the `inkview-slint` demo model).

## Verification

- Full engine workspace `cargo check` passes; `pocket-mod`/`pocket-ui-surface`/`pocket-widget` tests pass.
- Host compiles natively, is clippy-clean, and its framebuffer/input unit tests pass (6).

## Not done yet (draft)

- **On-device validation** — refresh tuning, touch accuracy, and the logical-viewport choice (the 9-bit touch coordinate limit forces a sub-512 logical viewport at density ≥2; see implementation guide §9). Needs hardware.
- **Target profile registration** — `pocketbook` entry in `contracts/spec/platforms.ts` + build backend (Phase 5). The host currently sets `set_identity("pocketbook", 4)` in anticipation.
- The `inkview` dependency is a local path dep for now; switch to the git dep in `hosts/pocketbook/Cargo.toml` for CI.


wonderbeel:
Apologies for the draft MR, I was trying to do some investigation/MVP integration work and didn't realize that after the final validations the agent wanted to open a MR (and it even got quite creative to bypass the fact that I cloned using HTTPS, big mistake installing `gh` on my machine...) :sweat_smile: 

Let me know if a e-ink host is still in scope of the project or not, I discovered it this morning and the first thing that come to my mind was "oh wow this is so much better than slint, gotta try it!", I will keep it local only in case (this still needs work and testing on a physical device of course but I own both a Era Color and a Verse so I can continue working on it).

wonderbeel:
It was a really fun experiment (was also a good way to bench the capabilities of Qwen 3.8) and surprisingly it works! I still need to properly test more features (and when I am back home I gotta test on my Era Color too) but the fact that it already renders without ghosting is not bad at all.
I have a small question about device targets, I went with a vita logical viewport target for compat to reduce the complexity of this experiment but of course it is not ideal, what would be the better approach for the future? Keep a compat target to make cross device development easier and make a new native target that has a dynamic viewport like desktop (actually would it be possible to use a dynamic viewport in takeover mode?) ?

photo_2026-07-24_09-42-49

https://github.com/user-attachments/assets/b6da5880-f5be-4ae5-9d8f-1b9ef6be4bf0

doodlewind:
Thanks for this awesome experiment! Will do some arch refactoring to make the host more scalable for landing this PR.

doodlewind:
Reviewed the whole PR and pushed three maintainer commits (a merge with current main + `31b1c04`-style cleanup + the incremental-raster rewire, `efd4cd2..e720e98`). This is a genuinely good host — the `pocket-ui-surface` extraction is exactly the right cut (mechanism moved verbatim, wgpu re-exports keep desktop consumers source-compatible), the refresh policy port is careful, and thank you for the Verse validation with photos.

## What I pushed

**Merge with main** — brings in #160/#173/#174/#155; everything below builds on #173's damage infrastructure.

**`refactor(pocketbook): align the host with repo conventions`**
- `contracts/spec/platforms.ts` + `tools/pocket.ts` had been wholesale re-indented (tabs); I restored main's formatting and reapplied only the pocketbook additions — the diff vs main is now pure additions. Please keep editor reformats out of shared files: they bury the semantic change and destroy blame.
- `tests/platform-contracts.test.ts` pins the target registry exhaustively (`["psp","vita","macos-widget"]`), so the full `bun run test` chain failed with the new target. Registered `pocketbook` + pinned its profile. Heads-up: this repo has **no CI for the bun chain** — run it locally before pushing.
- Moved the two repo-root `.md` design docs to `hosts/pocketbook/docs/` (`docs/STRUCTURE.md`: nothing else gets a top-level name), `deploy.sh` → `deploy.ts` (repo scripts are Bun TS; the device-side `.app` launcher stays `/bin/sh`), pinned `inkview` to a git rev so the crate builds standalone, and `input.rs` now reuses `pocketjs_core::spec::btn`.

**`feat(pocketbook): rasterize incrementally via core DrawList damage`**
#173 landed a backend-neutral `DamageTracker` in core; this host is its ideal consumer, and the two damage layers compose cleanly:
- `render_scaled_incremental` repaints only DrawList-diffed regions of the retained RGBA8 buffer — idle frames cost zero raster and zero scan (your full-frame diff was scanning 522k pixels every 33 ms tick even when nothing changed);
- your 16×16 tile diff now runs only *inside* those regions and still owns the e-ink refresh decision — it protects the panel from conservative damage that renders identical pixels;
- the retained buffer is always the complete frame, so Show/full re-blits without re-rasterizing; `prev` latches only blitted tiles.
Pixel parity with full renders is unit-tested (moving frames, density 2, structural changes, equal-pixel repaints). Host: 9 tests green, clippy clean; engine workspace, full bun chain, and 49/49 goldens verified locally.

## For your next device pass

The Verse validation predates the incremental rewire — please re-confirm animations + Show/background-return on hardware (the README checklist notes this), plus the still-open items (touch, hardware keys, idle ghosting, Era Color). One design question before undrafting: `hostAbi: 4` and the nominal `physicalViewport: [960, 544]` both look right to me, but flag anything the Verse pass contradicts.

doodlewind:
@wonderbeel could you double check if current refactored branch works you there? I think we can merge after your confirmation, thanks!

wonderbeel:
I will give it a try when I am back home, thank you for the extra pass 🙏

wonderbeel:
https://github.com/user-attachments/assets/0fa1681d-3069-4d60-99e8-6cc454bc951a

Tested after your refactor @doodlewind and it works just fine, the only thing that I noticed is that the progress bar animation now is slightly different but it actually looks more similar to how the demo renders on the desktop. If you are fine merging this branch as is I would flag that this host is still work in progress, I will try to iterate in the next weeks on top of it to complete it.

PS: If there is enough interest in e-ink devices my parents have a couple of jailbroken kindles, after pocketbook is stable I can try to tackle them too :).

doodlewind:
Merged — thanks for the device re-run and the video 🙏

The animation difference you noticed is expected and benign: the tick loop no longer re-rasterizes and re-scans the whole 960×544 frame every 33 ms, so the progress bar now runs at its intended cadence (hence the closer resemblance to the desktop host). Recorded that in the README.

Two things I changed while resolving the merge: #176 landed `symbian-e7-dev` on `hostAbi: 4` while this was open, so pocketbook took **5** (the check is on (target, abi) so nothing mis-bound, but the number is a registry-wide sequence) — the host constant moved with it. Also registered the target in `tests/symbian-runtime.test.ts`, a second exhaustive target-key assertion. The README now flags the host as work in progress with your open checklist items.

And yes — jailbroken Kindles would be very welcome. The e-ink layering here should port well: DrawList damage bounds the raster, the pixel tile diff decides what the panel refreshes, and only the panel-driver half (inkview → whatever the Kindle exposes) would be new. Happy to review whenever you get to it.

PR #189, opened 2026-07-26
fix(compiler): strip SFC template comments; parse comment templates as Comment nodes

odex21:
## Problem

Vue SFC template comments (`<!-- ... -->`) render as **visible text** on PocketJS.

Root cause chain:

1. The compiler keeps comments by default (dev default upstream; the cjs dist hardcodes `comments: true`), so they enter the template AST.
2. Vapor emits each kept comment as a static `template("<!-- ... -->") ` snippet.
3. The DOM facade's `parseTemplateHtml` cannot parse comment syntax (its tag regex requires a leading letter) and falls back to `createTextNode(html)` — a native text node painting the **literal comment source** on screen.

Kept comments also feed the pass-1 string collector, baking font-atlas glyphs no UI ever displays (**~750KB of .pak in a downstream lyrics app**), and materialize as useless nodes in the native retained tree (layout/raster walk, tree dumps, hit-testing).

## Fix

Three layers:

- **`vue-sfc-compile`**: strip comments at the SFC parse via `templateParseOptions`. `compileScript` hands the *pre-parsed* template AST to the Vapor compiler, so a `compilerOptions`-only setting arrives too late — verified by test. (`compilerOptions.comments: false` is kept as a backstop for compileScript-internal re-parses.) This matches upstream's own production default (`comments: false` when `NODE_ENV=production`); PocketJS has no SSR/hydration/devtools consumer of comments.
- **`vue-vapor-dom`**: `parseTemplateHtml("<!-- ... -->")` now yields a proper Comment node (invisible empty native text) instead of literal text — DOM-correct for comments arriving from any other path (hand-written `template()`, third-party precompiled components).
- **`jsx-plugin`**: hash the compiler's own sources (`jsx-plugin.ts`, `vue-sfc-compile.ts`) into the transform cache key. The key covered dependency versions and inputs but not this package's implementation, so a compiler-behavior change silently served stale cached output on warm caches (hit while developing this very fix). `CACHE_VERSION` stays as a manual backstop.

The JSX/Solid path is unaffected by construction: JSX has no HTML comment syntax (`{/* */}` is a JS comment dropped at parse), and the Solid preset compiles with `generate: "universal"` (direct renderer calls, no HTML template strings at runtime).

## Verification

- Two new regression tests: compiler strips comments from emitted code; shim parses comment templates as Comment nodes.
- Full `bun test` failure set identical to base commit (pre-existing environment failures only: mgba/switch/launcher harnesses).
- Downstream ESP32-P4 app rebuilt: zero comment templates in the bundle, live native tree drops the ever-present comment node (349 → 348), `bun web/verify.ts` + hit-ball + chaos suites pass, device boots with all stage checks PASS.

### DIFF
--- a/framework/compiler/jsx-plugin.ts
+++ b/framework/compiler/jsx-plugin.ts
@@ -79,20 +79,45 @@ const SOLID_UNIVERSAL_RUNTIME_PATH = new URL(
 const PACKAGE_NAME = "@pocketjs/framework";
 const CACHE_DIR = new URL("../../.cache/transforms/", import.meta.url).pathname;
 const CACHE_VERSION = "2"; // manual backstop; compiler sources are hashed in below
+const COMPILER_DIR = new URL("./", import.meta.url).pathname;
 
-// Transform behavior also depends on this package's own compiler sources,
-// which dependency versions and input hashes can't cover.
-const IMPLEMENTATION_SOURCES = ["./jsx-plugin.ts", "./vue-sfc-compile.ts"];
-let implementationHash: string | undefined;
-
-async function compilerImplementationHash(): Promise<string> {
-  if (implementationHash !== undefined) return implementationHash;
-  const h = new Bun.CryptoHasher("sha256");
-  for (const source of IMPLEMENTATION_SOURCES) {
-    h.update(await Bun.file(new URL(source, import.meta.url)).text());
+/**
+ * Hash of this package's own compiler sources: transform behavior lives here,
+ * and neither dependency versions nor input hashes cover it, so a warm cache
+ * would otherwise serve output from the previous implementation.
+ *
+ * The set is walked from this file's own imports rather than hand-listed - a
+ * literal list rots the moment a compiler module is added or split, which is
+ * exactly the failure this key exists to prevent. Only `framework/compiler` is
+ * in scope: everything under `framework/src` is a transform *input*, hashed as
+ * it is transformed.
+ */
+async function hashCompilerSources(): Promise<string> {
+  const scanner = new Bun.Transpiler({ loader: "ts" });
+  const sources = new Map<string, string>();
+  const pending = [new URL("./jsx-plugin.ts", import.meta.url)];
+  while (pending.length > 0) {
+    const url = pending.pop()!;
+    if (!url.pathname.startsWith(COMPILER_DIR)) continue;
+    const name = url.pathname.slice(COMPILER_DIR.length);
+    if (sources.has(name)) continue;
+    const source = await Bun.file(url).text();
+    sources.set(name, source);
+    for (const { path } of scanner.scanImports(source)) {
+      if (path.startsWith(".")) pending.push(new URL(path, url));
+    }
   }
-  return (implementationHash = h.digest("hex"));
+  const h = new Bun.CryptoHasher("sha256");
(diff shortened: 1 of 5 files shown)


doodlewind:
Reviewed end to end — the diagnosis holds and the fix is at the right layer. Merging with two follow-ups pushed as f81143a.

**Verified locally** (no CI runs on this repo, so this is the gate):

- Reproduced the bug on `main` across eight template shapes — root comment, comment nested in an element, `v-if`/`v-else` separator, comment-only child, comment-only template, inside `v-for`, inside slot content. Base emits `_template("<!-- hello -->", 2)` / `_template("<div><!-- inner --><span> ", 1)` for seven of them; all eight are clean on this branch.
- `templateParseOptions` really is the only layer that works: `parse()` spreads it into `compiler.parse`, and `compileScript` hands `sfc.template.ast` straight to `compileTemplate`, so nothing downstream re-parses in the normal path.
- The `compilerOptions.comments: false` "backstop" is not dead code, for a second reason worth recording: `isMultiRoot()` derives `preserveComments` from `parserOptions.comments !== false`, so the two settings have to agree or root-unit counting and the AST would disagree. It also covers the `inAST.transformed` re-parse branch, which does spread `compilerOptions`.
- No app or test source in the tree has a template comment, so there is no golden/tape churn from the strip.
- `bun run test`: 453 pass, 0 fail. `bunx tsc --noEmit`: clean.

**Follow-ups pushed (f81143a):**

1. `jsx-plugin` — the cache key now hashes the compiler sources reachable from `jsx-plugin.ts`'s own relative imports (scoped to `framework/compiler`, since `framework/src` files are transform *inputs* hashed as they are transformed) instead of the two-entry `IMPLEMENTATION_SOURCES` list. A hand-kept list reintroduces exactly the failure this key exists to prevent: split `vue-sfc-compile.ts` in two, or add a module, and warm caches silently serve stale output again. Validated by adding a depth-2 compiler module and editing it — all 29 entries re-keyed both times, and reverting restored the original keys. The memo also holds the promise now, so concurrent transforms read the sources once.
2. `vue-vapor-dom` — `parseTemplateHtml` parses whatever follows the comment rather than requiring the whole string to be one comment. The anchored `/^<!--([\s\S]*?)-->$/` mis-parsed `<!-- a --><!-- b -->` into a single comment whose data was `` a --><!-- b ``, and a comment ahead of an element fell all the way through to `createTextNode` — the failure mode this PR set out to remove. `indexOf` + recursion is also less code than the regex.
3. Tests cover both new parse shapes; the duplicated stub-host setup is now one helper.

**Left alone, for the record:**

- `parseTemplateHtml` still treats an element's inner content as raw text, so a comment *nested* inside one (`<div><!-- c --><span>`) is still stringified. That is the pre-existing shape of this minimal parser (nested elements do not survive it either), and after the compiler-level strip it is unreachable from SFC input — expanding the parser belongs in its own change.
- `framework/compiler/solid-plugin.ts` is dead code: nothing imports it, `jsx-plugin.ts` superseded it, and it carries its own `CACHE_VERSION = "7"` with the same stale-cache hazard you just fixed here — while `docs/DESIGN.md` and `site/content/docs/architecture.md` still describe it as the pass-1 transform. Worth a separate cleanup PR.

PR #191, opened 2026-07-26
feat(kindle): run Hero demo on Paperwhite 5

doodlewind:
## Summary

- add a production-admitted `kindle-pw5` target (309×412 logical pixels at 4× density, host ABI 5) with a standalone ARMv7 musl Rust host
- adapt the existing Solid Hero app for a portrait e-ink layout, direct touch press, static decoration, and high-quality post-interaction cleanup
- add pinned Kindle bootstrap, KUAL/USBNetwork helpers, key-only SSH, content-addressed deploy/reload tooling, and an SSH-backed DevTools mailbox
- make framebuffer, input, UI, powerd, child-process, and USB Mass Storage ownership fail closed and recoverable

## Why

PocketJS had early e-ink-facing pieces but no complete Kindle runtime or proven Hero path. This change carries the same Hero source through target resolution, native build, deployment, Gray8 framebuffer rendering, FBInk presentation, touch input, and safe device restoration on a Paperwhite 5.

The hardware pass also uncovered and fixes four device-specific failures:

- PW5 firmware reports controller-applied `rotate=3` even though the visible framebuffer is already upright
- the device Dropbear build rejects oversized SSH exec requests, so publish scripts now stream over stdin
- `powerd` may remain in `screenSaver` while the PocketJS framebuffer is visible, suppressing touch events; the launcher now records, verifies, and restores `preventScreenSaver`
- asynchronous FBInk helpers are drained or reaped before UI/power recovery, and the final refresh completes while the runtime launch gate is still held

## User/developer impact

`bun run kindle build|deploy|run|reload|dev` now provides a verified PW5 development loop. Runtime launch exclusively grabs the discovered evdev touchscreen, uses damage-aware DU updates followed by a flashing regional GC16 cleanup, refuses unsupported panel/input/power states, and restores the exact Kindle UI and power-management state on exit.

## Validation

- Paperwhite 5 / 11th gen, firmware 5.19.2: Hero rendered upright at 1236×1648; direct taps incremented Count; post-touch cleanup was visually accepted; shutdown released evdev and restored KPP/awesome plus `preventScreenSaver=0`
- final ARMv7 musl Hero + host build: `bun tools/kindle.ts build --plan=.pocket/kindle-pw5/plan.json`
- repository suite: `bun run test`
- Kindle TypeScript suites: 68 passed (`tests/kindle-bootstrap.test.ts`, `tests/kindle-tooling.test.ts`)
- Kindle Rust suite: 24 passed
- POSIX shell syntax checks and `git diff --check`


wonderbeel:
@doodlewind If you need help testing on other devices I have a [Kindle Oasis](https://wiki.mobileread.com/wiki/Kindle_Oasis) at home this week, I should also be able to fetch a [Kindle 5](https://wiki.mobileread.com/wiki/Kindle_5) if my parents haven't trashed it. 
I also have a quick code question, would it make sense to make a eink crate to share common logic? At a glance it seems that a lot of stuff could be shared between kindle and pocketbook, it would also make possible to add kobo and (maybe?) xteink devices but maybe it is better to postpone this standardization for when the devices are more stable.

PR #205, opened 2026-07-30
integrate Playdate target, host input, and example apps

colmugx:
IMG_0028


doodlewind:
**Verdict: fit to merge** — this is a model target addition, and it lands with one real rendering bug fixed forward on the branch (1440211). Reviewed with the repo review workflow (`skills/pocketjs-review-pr`).

## The finding: 1bpp polarity was inverted on device

`vp_pd_render_frame` wrote style 0 as raw glyph bytes and style 1 as `~glyph`. Font bits are **ink** (`vapor_esp32.c` pins this: `bits & 0x80>>x ? ink : paper`), and Playdate framebuffer bits are **white** — your own device photo proves it: the `bg-white text-black` title bar renders **white-on-black**, and the dark-theme body renders **dark-on-light**. Every pair displayed as the exact inverse of its declared luminance orientation, of this PR's own `styleTableCss` playdate preview (`s === 0 → ink #000 on #fff`), and of what GB/NES render for the same app under the shared `styles2` contract. It photographed fine because complementing every pixel preserves contrast — but it breaks the "same file, every target agrees" invariant.

Fixed to `style ? glyph : ~glyph` with the semantics stated in a comment, and the C harness now pins the corrected direction (`frame[0] == ~0x81` for style 0). **Note:** a rebuilt `.pdx` will look color-flipped relative to your photo — dark title text on a white bar, light body text on the dark theme — which is the orientation the oracle preview and DMG render. A fresh device glance when convenient would be welcome, but the direction is proven by the photo + the ink/white bit semantics, so I didn't hold the merge on it.

## Second gap: the oracle replay was claimed, not present

`DESIGN.md` says the Playdate Todo "is also replayed through the real Vue Vapor oracle with signed Primary-axis deltas" — no test booted `todo.playdate.tsx`; `entry.ts` hardcodes `todo.tsx`, so the new `axisDelta` hook had no listener. Rather than edit the claim down, I made it true: `bootOracle` takes a bundle entry, `entry-playdate.ts` mounts the crank variant, and three tests replay boot on the 50×30 grid, signed 45° detents with sub-detent accumulation and end-clamping, and edit-mode crank isolation. The grid-line assertions passed on first run — the app and oracle plumbing were already correct.

## What was verified as-is

- **Taxonomy**: Playdate slots into `VAPOR_TARGETS` as data; the BOARDS.md addition correctly argues why a fixed SDK target is *not* an ESP32-style board JSON. `relativeAxesUsed` joins `buttonsUsed` as a derived demand; VT101/VT102 admission errors extend the existing VB/VS-code convention without collision.
- **`RelativeAxis` contract**: hardware-neutral signed millidegrees, apps own detents — the runtime preserves sub-millidegree remainders across frames, drains docked/lifecycle motion (no ghost events), and the C runtime harness pins all of it against a fake SDK table. The dev server's wheel adapter uses the same accumulation. This is the right shape for the future ESP32 encoder, and BOARDS.md records that follow-up honestly.
- **Emitted-C changes for existing targets** (`condition()` paren trim, `app_on_axis_delta` stub, `VP_UNUSED_FN`): behavior-neutral, proven by the **GBA/GB/NES emulator per-press parity suites passing unchanged**. `esp32BuildId` shifts for identical sources as a consequence — that is the documented contract (the id follows generated code + runtime sources; the board-half pin in `boards.test.ts` is untouched), so a pre-#205 MeowBit will correctly report stale on `vapor:esp32:verify`.
- **Passing doc-rot fix**: the `screen.ts` "NES 24x20" comment was stale; 22×18 matches `VAPOR_TARGETS.nes`.
- `pd_api.h` in the test harness is a 62-line clean-room shim (function-pointer tables + enum values, no SDK expression). No new deps; `vapor/` stays outside the npm `files` map.
- CLAUDE.md: kept your rule, trimmed to the enforceable sentence — the design exposition lives in `DESIGN.md` §5.

## Local gate

- `bun run vapor:test`: **67 pass, 0 fail** (64 from the branch + 3 new oracle-replay tests; includes both C harnesses under `-Wall -Wextra -Werror` and the GBA/GB/NES parity replays)
- `bunx tsc --noEmit -p vapor` and root `bunx tsc --noEmit`: clean
- Root `bun run test`: **450 pass, 0 fail** (base-relative)
- Real SDK smoke on this machine (SDK 3.x at `~/Developer/PlaydateSDK`): simulator `.pdx` builds and validates end-to-end (51.3 KB). The device build needs the newlib-bearing ARM GNU toolchain that Panic's `arm.cmake` assumes — Homebrew's bare `arm-none-eabi-gcc` stops at `stdint.h` — so device evidence remains your on-hardware photo, which is exactly what `runtime/playdate/README.md` prescribes.

PR #227, opened 2026-08-05
perf(octane): 0.1.26 scoped updates, leaf state in hero, per-framework XMB covers

doodlewind:
Upstream shipped the fix for the press-latency issue we filed, so this bumps octane and adapts the one demo that shows what the fix is worth.

## What upstream did

[octanejs/octane#574](https://github.com/octanejs/octane/issues/574) (our report, with a standalone repro) was fixed the same day by [#578](https://github.com/octanejs/octane/pull/578) and released in **0.1.26**. A universal `setState` now replays only the owning component's subtree; full-root replay became the fallback. They also turned our repro into their permanent regression benchmark (`benchmarks/universal-leaf-update`, with a 4000/0 sibling ratio guard).

Measured with our repro on 0.1.25 → 0.1.26: a one-leaf `setState` beside 4,000 untouched siblings goes **22.96 ms → 0.010 ms**, and the curve is flat instead of linear in tree size.

## What that gave us on device: nothing, at first

The upgrade on its own is inert here. All 23 `psp-octane` goldens stay byte-exact, and PPSSPP press cost is unmoved on every bench app (solid control ±0%, so this is not engine drift from the main merge).

Probing the real hero bundle in the sim host showed why: every scheduling precondition passes (`depth 1, urgent, !fullRoot, owners 1`) but `resolveScopedTarget()` returns nothing. It walks up from the updated owner looking for one with a committed **range**, and the root component has none — so state held in a root component still replays the entire tree. All seven demos keep their interactive state in the root. The one exception proves it: `settings` is the only app with a nested `useState`, and it was the only app whose numbers moved.

## What this PR changes

`hero` adopts the shape the fix rewards. The counter moves into a `CounterRow` leaf, and the underline that the count also drove leaves the render path for the native `jump()` channel. A constant `translateX: 0` stays on the underline so its edge rounding matches the previous raster exactly.

| hero | avg frame work | worst frame (the press) |
|---|---:|---:|
| octane, root state | 6.53 ms | 176.89 ms |
| **octane, leaf state** | **3.64 ms** | **37.80 ms** |
| solid, same tape | 3.66 ms | 31.47 ms |

Press latency drops 4.7×, and octane lands at parity with Solid on this app. Rendering is unchanged to the byte.

## Not done here

The other six demos are unconverted. Several hold root state legitimately — `library`'s `screen` and `stats`' `tab` switch whole views, so a root replay is the honest cost there. Converting the rest deserves per-app judgement, not a blanket refactor.

Real-hardware confirmation on the PSP is still pending (the EBOOT is built; the memory stick was not connected during this run).

## Verification

- Full suite: 8 groups, 0 fail; `tsc --noEmit` clean
- 23/23 `psp-octane` PPSSPP goldens byte-exact, before and after the hero refactor
- Bench: 7 apps × {solid, octane} × 7 samples, 5000 bootstrap resamples; solid unchanged as control


doodlewind:
Added on top: per-framework XMB covers.

A demo's framework twins were the same icon and the same PARAM.SFO title, so a memory stick holding both the Solid and Octane build of `hero` gave no way to tell them apart. `gen-demo-covers.ts --framework=<fw>` now bakes a variant set into `apps/<demo>/psp/<fw>/` — framework name under the accent rule on the tile, an accent-bordered badge in the PIC1 corner, `(Octane)` on the SFO title, and PIC1 rendered from that framework's own bundle. `tools/psp.ts` prefers the variant directory when present, so default builds are unchanged.

All eight Octane demos now carry it, verified by unpacking the built PBPs rather than trusting the toml.

The eight release EBOOTs are on the memory stick (SHA-256 verified against the staged builds), replacing the previous Octane installs. Hardware confirmation of the 176.9 -> 37.8 ms press result is the last open item.

doodlewind:
Hardware confirmed on a real PSP — the press hitch is gone, matching the 176.9 -> 37.8 ms PPSSPP result, and the XMB now distinguishes the Octane twins by icon and title.

Merged latest main (v0.8.0, the new `tools/test.ts` stage runner) into the branch: 11/11 stages green, `tsc` clean, and all 23 `psp-octane` goldens still byte-exact.

Issue #59, opened 2026-07-07
Improve cross-platform abstractions for embedded Linux backends


haoyn231:
## Improve cross-platform abstractions for future embedded Linux backends

  Hi, I have recently been experimenting with running PocketJS on Linux/Wayland and made a few initial adaptations:
https://github.com/haoyn231/pocketjs/blob/feat/sdl2_linux_sim/pocket3d/examples/sdlhost/src/main.rs

<img width="1705" height="817" alt="Image" src="https://github.com/user-attachments/assets/a2273aed-3ac0-46d2-9ead-1bd9af93424f" />

  - Added a Linux SDL2 host to validate the PocketJS UI runtime on desktop Linux.
  - Added mouse/touch-style pointer input for the SDL host.
  - Fixed pointer coordinate mapping when SDL logical size is enabled.

  These changes are mainly meant to explore whether PocketJS can become practical on embedded Linux devices. My goal is not just to build a desktop SDL demo, but to eventually support common embedded Linux interfaces such as:

  - `evdev` input
  - DRM/KMS output
  - framebuffer/fbdev
  - Possibly a lighter software rendering or GLES path

  ### Background

  I think PocketJS could be very valuable for the embedded Linux ecosystem.

  Many embedded Linux devices need a more modern and composable UI development experience than LVGL, but cannot afford the cost of a full browser stack such as Chromium/WebView/Electron. PocketJS fits nicely into this space: UI and interaction can be described with Solid/Vue/JSX, while layout, styling, animation, text, and rendering are handled by a Rust/native core.

  This is attractive for low-memory, low-power devices that also need deterministic testing, such as:

  - Handhelds and game frontends
  - Industrial control panels
  - Camera, NVR, router, gateway, and appliance UIs
  - Factory test and debugging panels
  - Embedded products that need scriptable UI or plugin capabilities

  ### Current issues

  While adding the SDL2 host and pointer input, I felt that PocketJS is architecturally cross-platform, but parts of the current code are still strongly tied to PSP assumptions, fixed resolution, or specific hosts. This makes adding new backends require more changes in shared code than expected.

  Some areas that stood out:

  1. **Resolution and viewport abstraction**

     Many places still assume the PSP `480x272` resolution. That is reasonable for PSP, but embedded Linux devices may have very different screen sizes, orientation, and scaling requirements.

     It would be helpful to have a clearer viewport/display contract between the runtime, core, and host, for example:

     - Logical size
     - Physical size
     - Scale factor
[...]

PR #229, opened 2026-08-06
feat(input): implement input.pointer as a cursor position source

7086cmd:
## Why

The platform contracts have registered `input.pointer` since the desktop surface landed — *"a REAL absolute pointer (mouse/trackpad): position plus press/drag/release edges, hover resolves focus"*, explicitly **not** `input.cursor` ([`contracts/spec/platforms.ts`](https://github.com/pocket-stack/pocketjs/blob/main/contracts/spec/platforms.ts), `docs/WIDGET.md`). Nothing in the framework consumed it.

The only worked example, `examples/note-widget`, routes its mouse through an app-specific `svc` channel and maps clicks to CIRCLE by hand:

```rust
// Clicks are CIRCLE — hover already focused what's under the pointer.
let buttons = if mouse_down { BTN_CIRCLE } else { 0 };
```

So every host with a real pointer reinvents hover and hit-resolution in app code. This gives the capability a framework implementation.

## What

```js
enableCursor({ source: "pointer" })
```

The host's absolute position drives the **existing** cursor state machine and the pointer's own button supplies the press/release edges. Hover, `focus:`, `active:`, press arming and `onPress` behave exactly as they already do — only the position source and the edge source change. `speed`, `dpadSpeed` and `button` do not apply.

Default stays `"analog"`, so every existing app, host, tape and golden is untouched.

Position arrives as a 4th frame argument, packed `(down << 20) | (y << 10) | x` (10 bits per axis), mirroring how `touches` was added. `undefined` means the host has no pointer this frame — a pointer that left the window — deliberately distinguishable from one parked at the origin: the cursor holds its last position and an in-flight press stays armed until a real release.

`pocket-mod` gains `Guest::frame_with_pointer` for native hosts.

### Why a frame argument, not a polled `HostOps` method

A pull-based op would not be captured by the flight recorder, and every input the guest sees has to be on the tape for `tape:check` to mean anything. Replay scrubs the pointer exactly like touch.

## Second commit: `apps/hero`

Hero marks its button `focusable` with `onPress` but never establishes initial focus, and the framework deliberately does not auto-focus — `firePress` walks up from `focused`, which is null until a d-pad press. So CIRCLE on a freshly mounted hero activates nothing, and a host reporting an absolute point (Vita/PocketBook panel, or a desktop host presenting a mouse as a contact) could not reach the button at all.

It now resolves the first contact to the focusable under it in `onFrame`, which runs before input edge detection — the same hover-IS-focus pattern `apps/note` already uses, spelled against the portable touch channel. Hosts with neither touch nor a pointer report no contact and it is a no-op.

## Verification

- `bunx tsc --noEmit` clean
- `bun run tape:check` — **180 frames still match**, both commits
- 7 new cursor tests for the pointer source: position adoption vs. nub steering, hover-focus with no button held, press/hold/release firing once, cancel on release away from the armed node, the PSP mask **not** clicking a host-driven cursor, pointer-left-the-window keeping a press armed, viewport clamping
- 5 new `pointer.ts` snapshot tests, registered in `package.json`'s `test` script
- `pocket-mod` round-trip test for `frame_with_pointer`
- Full `bun run test`: 242 pass, 1 fail — `psp-toolchain.test.ts > doctor treats an invalid explicit LLVM override`, which imports only `tools/psp-toolchain.ts`, has zero references to `framework/` or `pocket-mod`, and fails on a machine with no PSP SDK. Please confirm it's green in your CI.

## Context

Found while wiring mouse input for a `consortium-hmi-pocket` kiosk host (embedded Linux, where there is no OS cursor to fall back on). Happy to adjust the packing, the option name, or split the hero commit out.


doodlewind:
Thanks for working on this. The underlying goal is valid: give a real mouse or trackpad pointer a framework-owned path for absolute position, hover-to-focus, press, drag, release, and onPress, instead of making each host send private svc mouse messages and synthesize CIRCLE.

I am closing this PR because the implementation is based on an input model that current main has superseded, and carrying it forward would lock in the wrong boundaries:

- frame argument 4 now carries touch hit facts, so the proposed pointer frame ABI conflicts with current main;
- pointer input is scrubbed during replay but is not actually recorded or restored by the tape;
- the 10-bit coordinates contradict the macos-widget 4096 by 4096 dynamic viewport, and the cached cursor viewport does not follow live resize;
- a single down level cannot preserve a fast press-and-release in one tick or express leave, blur, and cancellation safely;
- input.pointer is coupled to the virtual input.cursor sprite operations;
- no stock host uses the new path, so there is no end-to-end host receipt;
- the Hero change consumes touch rather than pointer and is superseded by the framework-level touch activation now on main.

This is a close-and-rework decision, not a rejection of mouse support. Please coordinate the next pointer implementation in #59 and start from current main. A focused replacement should define one versioned frame-input contract, explicit pointer edge and cancellation semantics, tape recording and replay, separation between real-pointer interaction and virtual-cursor rendering, and one real host integration such as note-widget or Linux SDL. It should also cover Solid, Vue Vapor, and Octane, plus resize, high-resolution, fast-click, cancellation, and replay tests.

Thanks again for pushing this direction forward.

PR #235, opened 2026-08-06
perf(iphone2g): scope the composite to damage, and reach a locked 60 fps

doodlewind:
0.9.2.

**The composite was the only stage that was never damage-limited.** `pocket_draw_rect` opened with `(void)rect;`, discarding the dirty rectangle UIKit hands it, and the tick invalidated the whole view — so every frame rebuilt a `CGImage` over all 320×480. That was 22–27 ms.

An empty damage plan now invalidates nothing (626 of 961 frames in one sample); a non-empty one goes to `setNeedsDisplayInRect:` and `drawRect:` clips. **Composite 22–27 ms → 0.26 ms. Frame 33.9–40.8 ms → 7.63 ms. 21.9–26.5 fps → 59.99.**

The software rasterizer becomes the default and GL becomes opt-in. This **supersedes rather than corrects** 0.9.1 — the GPU path really was faster as that code stood; the CPU path then got an optimization GL still lacks.

**Damage is now observable**, which is how this was found: `engine/symbian` was discarding the `DamagePlan`. Six C ABI accessors plus a `composites` counter now ride in the device record. They immediately refuted the hypothesis they were built to confirm — damage was working fine, and the composite was throwing it away.

**Verified:** software capture diffed against the reference core after 2,581 damage-limited frames — mean 0.039/255, residue confined to the animating spinner's 40×40 box, `damage_failures=0`.

Gate: 11/11 stages, 54/54 goldens, 180 tape frames, engine/core 110 tests, both GL pipelines 0 warnings, tsc clean.


doodlewind:
Superseded by a clean branch off current main — this one reused a branch that had already been squash-merged, so it conflicted with itself.

PR #238, opened 2026-08-06
feat(fs): the fs module — a per-app file tree behind a nine-op spec, sim host, pocket-fs reference core

siwei-yuan:
> **Stacked on #231** (the db module) — review the last commit pair; the first two are #231's. Will rebase to a clean diff once #231 merges.

The fifth module-shaped vertical slice (after `ui`, `strike`, `audio` and db), built spec-first per the module discipline: a per-app file tree mounted as `globalThis.fs`, capability id `data.fs`.

## The boundary

`contracts/spec/fs.ts` pins nine synchronous ops — `read` / `write` / `remove` / `list` / `stat` / `mkdir` / `rename` / `usage` / `lastError` — plus:

- **universal names**: a segment is any well-formed Unicode — Chinese names, dot-prefixed names, spaces — except `.` and `..` (the escape hatches), control characters, and oversize (64 UTF-8 bytes / segment, 8 segments, 160 bytes / path). **No name is reserved to the host.** Names are byte-for-byte identities (no case folding, no normalization) — the folding-filesystem caveat is documented, and the deterministic hosts are byte-exact so goldens catch collisions early;
- **isolation by construction, independent of names**: every path is relative and resolves under the root the host binds at mount; `..`, absolute paths, and `/`-in-name are unrepresentable, so apps cannot spell each other's trees — the ATTACH-refusal principle generalized. The reference core lstat-refuses symlinks planted by host-side actors. Privilege is the binding: on Pocket Pi the device agent is the same module bound at `/workspace`, apps at `/workspace/apps/<id>/data/`;
- **payload encoding**: text as a JSON string (stored as UTF-8), bytes as the db module's `{"$b": base64}` spelling; 64 KiB per op crossing (the SDK chunks larger files transparently);
- **atomicity with a clean tree**: truncate writes land in the module's own temp directory — outside the bound root, same filesystem — then rename over the target (the power-loss contract LittleFS hosts inherit). The module owns `tmp` and clears it on construction, so the app's tree never shows host machinery and a crash orphan cannot outlive the next boot;
- **paged list**: sorted by Unicode code point (= UTF-8 byte order; the sim host carries the comparator since JS sorts UTF-16 code units), `FS_MAX_DIR_ENTRIES` (256) per call with offset + eof;
- **frame contract**: no clock, no events, and `stat` carries **no mtime** — a timestamp is the fs spelling of `Date.now`, excluded for the same golden-test reason. No `watch()`.

One data root serves both data modules: a database is an ordinary file (`<root>/<name>.sqlite`) in the app's home — its own asset, visible like any of its files (backup = a file copy). Overwriting it corrupts the app's own data, the same trust class as deleting its own files; SQLite fails loudly on a corrupt image.

## What ships

- **`@pocketjs/framework/fs`** — the Bun shape: `file()` (`.text/.bytes/.json/.size/.exists`) and `write()`, plus the node:fs sync subset Bun implements (`readFileSync`, `writeFileSync`, `appendFileSync`, `mkdirSync`, `readdirSync`, `rmSync`, `renameSync`, `statSync`, `existsSync`), and `usage()`. Everything returns synchronously per the frame contract, and `await` unwraps a plain value, so Bun-idiomatic code (`await Bun.file(p).text()`) migrates unchanged. Throws where the namespace is unmounted, like db: file code that silently drops writes is a corruption bug.
- **`framework/src/bytes.ts`** — the base64 codec extracted from db-api plus a strict UTF-8 decoder (QuickJS has no TextDecoder), shared by both data SDKs.
- **`hosts/sim/fs.ts`** — in-memory tree behind the op namespace, injected via `bootWorld` `extraGlobals`; byte-exact names, code-point-sorted listings.
- **`engine/crates/pocket-fs`** — the reference core: `Storage::Memory` / `Storage::Dir { root, tmp }` over `std::fs`, root confinement, per-segment lstat symlink refusal, atomic truncate writes through the host-owned temp dir, per-app quota. `mount` is a default feature; `default-features = false` drops the pocket-mod dependency for firmware with its own QuickJS wiring — verified to `cargo check` clean for `riscv32imafc-esp-espidf`.
- **`hosts/esp32p4/examples/data-smoke`** — the ppa-smoke pattern's second instance: both module cores driven on a real ESP32-P4 over LittleFS (see the commit for measured numbers; persistence proven across resets, including SQLite transaction atomicity through interrupted runs — the sample count stays a multiple of 288).
- **`docs/FS.md`** — the boundary, the `data/` + `tmp/` layout, and the three-move adoption path.
- **`data.fs`** registered ahead of any stock TARGET advertising it — the `audio.pcm` / `data.sqlite` precedent: the sim host and the reference core implement and test the whole contract.

## Verified

- `bun run test` 11/11 stages green (`tests/fs.test.ts` 15 pass: the op contract and the SDK; suite 331+ pass, contract byte-compare included)
- `cargo test -p pocket-fs` 12/12, including a live QuickJS guest round-trip, universal-name round-trips, orphan-sweep, and the symlink-escape refusal
- `cargo check --workspace` clean, clippy clean
- `cargo check --target riscv32imafc-esp-espidf` (no default features) clean
- On-device: `DATA-SMOKE: PASS boot=1` / `PASS boot=2` on a Waveshare ESP32-P4 rev 1.3 (fs contract ~0.45 s, 288-row tx ~0.4–0.6 s)


siwei-yuan:
On-device verification added: `hosts/esp32p4/examples/data-smoke` (the ppa-smoke pattern — one building block, one hardware verifier) runs both module cores with `default-features = false` against a LittleFS partition on a Waveshare ESP32-P4 rev 1.3:

```
DATA-SMOKE: fs ok in 505.072ms; boot 1; usedBytes 12299
DATA-SMOKE: db ok in 1.044217s (288-row tx 538.36ms)
DATA-SMOKE: heap before 582008 after 580780 (delta 1228)
DATA-SMOKE: PASS boot=1
```

`PASS boot=2` after another reset — a boot counter (fs truncate-write) and a per-boot 288-row transaction persist across resets, verified as boots × 288 on every start. The traversal/dot-name refusals, the ATTACH refusal, and the `.db/` fence location all hold on-device.

siwei-yuan:
Design revision after review (PR body updated to match):

1. **Names are now universal.** The earlier filename-safe-token grammar over-restricted: apps could not name their own files in Chinese or with a leading dot, and the restriction was never load-bearing for isolation (that's the binding + unrepresentable `..`/absolute paths + symlink refusal, unchanged). A segment is now any well-formed Unicode except `.`/`..`, control chars, and oversize. Nothing is reserved to the host.
2. **Atomic-write temps moved out of the app's tree.** `Storage::Dir { root, tmp }`: temps land in a host-owned directory outside the bound root (same filesystem — rename stays atomic), which the module clears on construction. The app's tree provably holds only the app's own names, and a crash orphan cannot outlive the next boot — a guarantee the old in-tree dot-file scheme could never give once dot names became legal.
3. **The `.db/` fence is gone** (with #231's update): a database is an ordinary visible file in the shared data root. The fence protected an app from itself — an app that can already `remove("", recursive)` its whole home gains nothing from one file being unspellable, and visibility makes backup a file copy.

Re-verified end to end after the revision: full suite green, clippy clean, espidf cross-compile clean, and on-device `PASS boot=1` / `PASS boot=2` — including universal names (CJK + dot-prefixed) round-tripping on real LittleFS, and SQLite transaction atomicity witnessed across interrupted runs (sample count stays a multiple of 288 through resets).

doodlewind (approved):
**Verdict: approved.** Same quality bar as #231 — the nine-op boundary is clean, the atomic-write contract is carefully built (host-owned temp outside the root, swept on construction), and the Dir backend's symlink refusal is real. I ran the same op matrix through the sim and the Memory core, found the issues below, and pushed the fixes as `a17c02c`. **This branch is rebased onto #231's fixed head**, so it will show a clean diff once #231 lands.

## What the review found (and the commit fixes)

1. **`tests/fs.test.ts` was never wired into the gate.** The PR body claims 331+ passing, but `bun run test`'s unit stage is a hand-maintained list in `tools/test.ts` and this file's entry was missing — the gate actually ran 316. Wired in; the stage now runs 331. (main's `test-suite.test.ts` guard only checks `iphone2g-*` files — generalizing it is the follow-up below.)
2. **`fsValidSegment` accepted unpaired surrogates**, contradicting the spec's "any well-formed Unicode". A lone surrogate has no UTF-8 spelling: a JS host stores it byte-exactly while the QuickJS→native bridge mangles it into a *different* name — silent identity divergence between sim and a native core, exactly the class of bug the byte-exact hosts exist to catch. The shared predicate now refuses it (pinned in tests), and the spec documents that a text *payload* with a lone surrogate is likewise host-dependent, so arbitrary bytes belong in the `{"$b"}` spelling.
3. **sim `list()` mishandled a negative offset** — passed straight to `Array.slice`, which wraps to slice-from-the-end, where the reference core clamps to 0. Clamped to match. A cross-host parity probe now byte-agrees on the whole op matrix (modulo JSON key order), including that astral-plane names sort by code point on both hosts.
4. **docs/FS.md now states the chunking atomicity caveat.** The op is the atomic unit, so an SDK write above `FS_MAX_IO_BYTES` crosses as truncate + appends and power loss between chunks can leave the leading chunks only — whole-file atomicity above 64 KiB is write-a-sibling + `rename`, the same move the module itself makes. The single-op atomicity contract was well-documented; this closes the multi-op gap.
5. **concepts.md's module diagram** enumerated ui/audio/strike; db and fs join it. Both PRs call themselves the 4th and 5th modules, but neither updated the page that enumerates them.
6. Minor: dropped a dead `pad` counter in `bytes.ts` carried over from the db-api extraction.

Verified consistent, left as-is: the Dir backend's lstat-per-segment symlink refusal (the security-critical path) round-trips correctly and hides planted links from `read`/`stat`/`list`; universal names (Chinese, dot-prefixed, spaces) round-trip byte-exact; the quota accounting agrees between hosts.

## Verified

- `cargo test -p pocket-fs` 12/12 (incl. the Dir-storage atomicity + symlink-escape test), clippy clean, `cargo check --workspace` clean
- `tests/fs.test.ts` + `tests/db.test.ts` 33/33; unit stage 330/331 — the 1 fail is the **pre-existing** symbian-runtime Gatekeeper flake described in #231, unrelated to this diff
- `bunx tsc --noEmit` clean
- The on-device data-smoke (both module cores on a real ESP32-P4 over LittleFS, persistence across resets incl. SQLite tx atomicity) is taken as hardware-verified per the ppa-smoke precedent.

## Left alone / follow-ups (not this PR)

- Generalize `test-suite.test.ts` to require every `tests/*.test.ts` to have a stage home (with an explicit exclusion list), so the next module can't repeat finding #1.
- The symbian-runtime 5s-timeout flake on managed macOS hosts (shared with #231).

doodlewind (approved):
Thanks for the feat!

doodlewind:
**Verdict: merge.** The 2026-08-07 review findings are already folded in as the branch's fix commit (fs tests wired into the gate — the suite the PR claimed was never running them; lone-surrogate segment names refused by the shared predicate; sim `list()` negative-offset clamp aligned with the core's `max(0)`).

What changed since: rebased over `main`, which now carries the net module (#246) and the merged db module (#231). Conflicts were union-appends in `gen-rust.ts` (fs import block + generator section ordered db → fs → net) and `tools/test.ts`; `engine/core/src/spec.rs` was regenerated from the resolved generator, and `site/content/docs/concepts.md` needed a semantic merge — the module diagram now lists all six modules (`ui · audio · db · fs · net · strike`) and the SDK example sentence names the db and fs shapes so the prose matches the diagram it sits next to.

Verified on the rebased head:
- `bun run test` — **11/11 stages green** (one unit-stage run tripped this machine's known Gatekeeper first-launch stall on freshly linked test binaries — the affected files pass in isolation and the full re-run is green; the flake predates this PR)
- `bunx tsc --noEmit` — clean
- `cargo test -p pocket-fs` — **12/12**, including the live QuickJS guest round-trip and the symlink-escape refusal
- `cargo check -p pocket-fs --no-default-features` — clean (the pocket-mod-free build for firmware with its own QuickJS wiring; the `riscv32imafc-esp-espidf` build plus the on-device `DATA-SMOKE: PASS boot=1/2` runs were verified on the Waveshare ESP32-P4 already)
- `cargo check --workspace` — clean
- `bun run gen` / `gen-rust.ts` — both idempotent on the committed files

Design conformance, for the record: spec-first nine-op boundary with universal byte-identity names and isolation by construction (`..`/absolute/`/`-in-name unrepresentable), atomic truncate writes through the module-owned temp dir, `data.fs` registered per the orphan-capability precedent, sim host injected via `bootWorld` extraGlobals, one subpaths row + generated exports, `framework/src/bytes.ts` as the shared codec home for the data SDKs (the net module's hand-rolled UTF-8 codecs will fold into it as a follow-up).

PR #256, opened 2026-08-09
feat(ios): pocket ios — transitional ios-dev target, NativeScript shell, play on the simulator

NathanWalker:
**Stacked on #255** (`feat/apple-host`) — the first six commits here are that PR; review the last six. Rebases onto main once #255 merges.

## What

The follow-up #255's description names: a dev-profile target id and CLI integration for the Apple host. `pocket play ios nsengine` now builds a guest against a resolved plan, stages it into a committed NativeScript shell, and launches it on an arm64 iOS simulator.

- **`tools/ios-profile.ts`** — transitional `ios-dev` profile (hostAbi 7, platform `ios`, form `embedded`, 480×272, raster density 1–4 as a registry input, capabilities `input.touch` + `text.glyphs.baked`), following the `tools/iphone2g-profile.ts` pattern: out of `POCKET_TARGETS` until device-level acceptance.
- **`tools/ios.ts`** — `doctor` / `setup` / `devices` / `native` / `build` / `stage` / `play`, mirroring the symbian shape. `pocket play ios <app>` delegates here. Doctor is green-path only on Apple Silicon (both native artifacts ship arm64 slices only); Rust is needed only for `--rebuild-native` — the published plugin carries a prebuilt xcframework. CocoaPods is verifiably not required.
- **`hosts/apple/ns-shell/`** — a minimal committed NativeScript app, plan-driven (viewport, density, and guest mode come from the staged plan + `current.json`), answering the `ns.ping` service channel. Runs in place for incremental rebuilds.
- **`apps/nsengine/pocket.json`** — the demo's first committed manifest (the root-template fallback demands `input.buttons`, which this surface cannot advertise).
- **Identity fix in `PocketSurfaceView.m`** — plan-built bundles bake `__POCKET_TARGET__`/`__POCKET_HOST_ABI__` and `assertNativeHostContract` refuses hosts that don't publish the pair. The surface now applies `pocket_apple_set_identity("ios-dev", 7)` at init (the C ABI existed; nothing called it). External-guest hosts publish the same pair on the `ui` namespace they mount ([`@nativescript/pocketjs`](https://github.com/NativeScript/pocketjs)). Without this, the very first plan-built run fails at mount.
- Tests: `tests/ios-profile.test.ts` (profile privacy, exact plan, density selection, refusals, an identity source-text guard, `checkAppTypes`), the `nsengine` admission-matrix row, CLI dispatch cases.
- Docs: `docs/APPLE.md` runbook, `docs/STRUCTURE.md`, README platform-evidence row.

## Validation (iOS 26.5 simulator, Apple Silicon)

- `pocket ios doctor` fully green; `pocket play ios nsengine --density=4` from a wiped shell (**cold ~47 s**) and warm (**~16 s**): nsengine renders, spinner animates, and the guest→host stat reads `pong 1` with no interaction — the whole `runEffect → svcSend → onEffect → post → svcPoll → deliver` loop through the shell's handler.
- `--external-guest`: platform stat reads `iOS 26.5 via NativeScript` (guest code reading `UIDevice` through the metadata bindings). Same bundle, both modes.
- `bun tools/test.ts` unit stage green including the new file.

## Notes for review

- The shell's `package.json` pins `@nativescript/pocketjs` and `@nativescript/ios-quickjs` by their npm names; both publish imminently. Validation above used `--plugin-path`/`--runtime-tgz` (documented in `docs/APPLE.md`), which override for the install and restore the committed template.
- No `POCKET_TARGETS` change and no schema change — promotion is a follow-up gated on a device acceptance suite, per the transitional-profile convention.
- **ProMotion follow-up:** PocketSurfaceView pins its CADisplayLink to 60 because the core advances in exact 1/60 s steps (the determinism/baked-motion contract). Driving iPhone ProMotion displays at 120 needs the step rate to become a target-profile property (durations resolved against the profile tick, `CADisableMinimumFrameDurationOnPhone`, and an adaptive `preferredFrameRateRange`). Note the iOS **simulator** reports `maximumFramesPerSecond = 60` regardless, so 120 is device-only evidence.


doodlewind:
## Review — one hard blocker (npm), four fixes pushed, follow-ups listed

**Verdict: blocked on a single external fact — the shell pins `@nativescript/ios-quickjs@9.0.0-preview.10`, and npm has nothing newer than `9.0.0-preview.3`** (checked 2026-08-10; `@nativescript/pocketjs@0.1.0` IS published, and I verified its tarball really carries the prebuilt `PocketApple.xcframework`, arm64 + arm64-sim). Until preview.10 publishes or the pin moves to an existing version, a cold `pocket play ios` fails at `npm install` for everyone. Everything else in the PR held up under review or is fixed on the branch — with the pushed commit the full gate is green, so this merges as soon as the runtime package exists.

One more publish-coordination fact for the stack: `@nativescript/pocketjs@0.1.0` was published ~90 minutes **before** #257's commit, so its headers, `index.js`, and `index.d.ts` contain no `tickRate` anywhere (verified in the tarball). The shell's `pocket.tickRate = staged.tickHz ?? 60` (from #257) is a silent no-op property write against that plugin — fine at 60, silently wrong at `--hz=120`. The plugin needs a republish once #257's ABI lands.

### What I ran

- `bun run test` with the pushed fixes: **11/11 stages green in 40 s**. Before the fixes the `launcher sim` stage failed (2/18 — measured, and bisected to this branch: base `9c809bb` runs 18/18).
- `pocket ios doctor` against a machine with no simulator runtime, no device, no ns CLI: all three reported accurately, everything else green — the doctor is honest.
- `bunx tsc --noEmit` clean; `tests/{ios-profile,cli,npm-package,test-suite,launcher-sim}.test.ts` individually green.
- npm registry + tarball inspection as above.

### Pushed to this branch (f5a5975)

1. **Launcher display union regenerated.** `apps/nsengine` is the repo's first Vita-only demo (PSP advertises no `input.touch`), so the committed `registry.generated.ts`/`images.json` were stale and `launcher sim` failed. Also evolved the Vita-equality pin in `tests/launcher-sim.test.ts` into what it actually guards: PSP's deck ⊆ Vita's, with the Vita-only delta asserted exactly. Covers are gitignored build artifacts (`bun tools/launcher.ts covers` renders the new one), so nothing else to commit.
2. **npm files map.** `tools/cli/README.md` ships `pocket ios …` to `@pocketjs/cli` users, but the governed files map carried none of `hosts/apple`, `apps/nsengine`, `docs/APPLE.md` — a published install failed at the shell/manifest lookup. Added the three entries following the iphone2g precedent + the pin in `tests/npm-package.test.ts`. `engine/apple` stays git-only: the default flow consumes the prebuilt plugin.
3. **`spawn` drain.** The copy dropped `tools/symbian.ts`'s concurrent pipe drain and timeout: awaiting `child.exited` before reading the pipes deadlocks once `simctl list -j` outruns the 64 KB pipe buffer. Adopted the symbian shape (`Promise.all` + optional `timeoutMs` → 124).
4. **Test-family governance.** `tests/test-suite.test.ts` only enforced unit-stage registration for `iphone2g-*.test.ts`; added the same guard for `ios-*.test.ts` so a future ios test can't be committed and never run.

### Follow-ups (not pushed — they need decisions)

- **`installShellDependencies` rewrites the committed `hosts/apple/ns-shell/package.json` with machine-local absolute `file:` paths**, restored only in `finally` (`tools/ios.ts:318-348` as committed) — an interrupted run leaves a tracked file dirty with a local path, and two concurrent runs clobber each other. Suggest never mutating the template: `npm install <pkg>@file:… --no-save`, or staging a copy under the already-ignored `.pocket/ios/`.
- **Reproducibility of the shell**: `^0.1.0` caret range on the one package that carries the xcframework **and** publishes the `ios-dev`/7 identity leg; `package-lock.json` gitignored; and install is skipped whenever `node_modules` exists, so a committed version bump never takes effect on an existing checkout. The repo's toolchain idiom pins external inputs by revision + hash in a data file (`tools/cli/*-toolchain.json`).
- **Identity in the wrong layer**: `PocketSurfaceView.m:14-16` hardcodes `"ios-dev"`/7 — a tools-owned transitional profile id baked into a reusable engine-layer UIKit class. The source-text guard in `tests/ios-profile.test.ts:98-105` does pin the two literals together (good), but a second Apple host inherits `ios-dev`/7 silently; an init parameter set by the profile owner would put it where density already lives.
- `--no-build` reuses `.pocket/ios/<demo>.plan.json` with no stamp check and silently discards `--density` — `tools/play.ts:177-206` implements exactly the missing stamp for vita.
- `pocket play ios` spawns a third process (`bin.mjs` → `play.ts` → `ios.ts`) though `iosMain(args)` is exported for in-process dispatch, and the branch sits before `demos()`/unknown-flag validation, so `pocket play ios nosuchdemo` reports a manifest-resolution error instead of `unknown demo`.
- `vendPluginXcframework` does `rmSync(..., {recursive: true})` inside a user-supplied `--plugin-path` with no shape assertion on the target (`tools/ios.ts:351-360`), and runs on every stage/play with the flag, not only `--rebuild-native`.
- `current.json` is a hand-written cross-boundary schema (writer `tools/ios.ts:306-316`, reader `ns-shell/src/app.ts:9`) with no shared type and no test, while the plan half of the same handshake goes through `extractHostBuildInputs` with a `planHash`.
- Smaller: runtime version compared via `parseFloat` (`tools/ios.ts:98`); `docs/STRUCTURE.md`'s `engine/` list doesn't name `engine/apple` (and the `crates/` line predates pocket-db/fs/net); doctor never checks `npm` or that the shell dir exists; `docs/APPLE.md` mostly reads in-register — the bolded labels at `:55`/`:69` bold the framing rather than the fact, and ":84 “Guest code reaches the whole iOS platform with no per-API glue”" is a one-liner the docs rule excludes.

Verified clean, for the record: hostAbi 7 is the next free integer (1 psp, 2 vita, 3 macos-widget/embedded, 4 symbian-e7-dev, 5 pocketbook, 6 iphone2g-dev — nothing enumerates the sequence though, worth a registry note someday); the transitional profile stays out of `POCKET_TARGETS` and both target-key pins still pass; the shell is plan-driven as claimed (no viewport/density/identity constant duplicated into shell source); `.gitignore` additions are correct and the `file:`-symlinked plugin tree cannot be swept in by `git add -A`; the nsengine manifest's stated rationale (root-template fallback demands `input.buttons`) checks out against `tools/demo-identity.ts:35-42`.

doodlewind:
Merged with the branch-merge of main (conflict resolution kept the review-fixed engine/apple files and this PR's PocketSurfaceView identity fix) plus @NathanWalker's ios-quickjs pin fix cherry-picked from feat/iphone16-demo — the pin belongs to this PR's file, so it lands here rather than at the stack top. Full local gate on the merged branch: 11/11 stages green.

Two things still owed on the plugin side, both noted in the review above: a cold-run confirmation of `pocket play ios nsengine` against the published `@nativescript/pocketjs@0.1.0` + `ios-quickjs@9.0.0-preview.3` (this machine has no simulator runtime to verify with), and a plugin republish carrying the tick-rate ABI once #257 lands.

PR #258, opened 2026-08-10
feat(apps): iphone16-demo 

NathanWalker:
Stacked on #257 (feat/tick-rate); only the last commit is new.

Following the `apps/iphone2g-demo` pattern, this adds `apps/iphone16-demo`.

Validated: `bun tools/ios.ts build iphone16-demo --density=4 --hz=120` resolves the ios-dev plan and builds; `bunx tsc --noEmit` and `bun test tests/ios-profile.test.ts` pass.

### DIFF
--- a/apps/iphone16-demo/app.tsx
+++ b/apps/iphone16-demo/app.tsx
@@ -1,0 +1,13 @@
+import Hero from "../hero/app.tsx";
+import { TICKS_PER_SECOND } from "@pocketjs/framework/clock";
+
+export default function IPhone16Hero() {
+  return (
+    <Hero
+      actionLabel="Tap Me"
+      deviceLabel="running on an iPhone 16 Pro, iOS 26.5."
+      presentationHz={TICKS_PER_SECOND}
+      runtimeLabel="RUST + QUICKJS + UIKIT"
+    />
+  );
+}
--- a/apps/iphone16-demo/main.tsx
+++ b/apps/iphone16-demo/main.tsx
@@ -1,0 +1,5 @@
+// @title PocketJS: iPhone 16 Hero
+import IPhone16Hero from "./app.tsx";
+import { mount } from "@pocketjs/framework/solid";
+
+mount(() => <IPhone16Hero />);
--- a/apps/iphone16-demo/pocket.json
+++ b/apps/iphone16-demo/pocket.json
@@ -1,0 +1,24 @@
+{
+  "$schema": "https://pocketjs.dev/schema/pocket-2.json",
+  "pocket": 2,
+  "id": "dev.pocket-stack.iphone16-demo",
+  "name": "pocketjs-iphone16-hero",
+  "title": "PocketJS: iPhone 16 Hero",
+  "version": "0.1.0",
+  "engine": {
+    "capabilities": {
+      "requires": ["input.touch", "text.glyphs.baked"]
+    }
+  },
+  "app": {
+    "entry": "apps/iphone16-demo/main.tsx",
+    "output": "iphone16-demo-main",
+    "framework": "solid",
+    "viewport": {
+      "fixed": {
+        "logical": [480, 272],
+        "presentation": "integer-fit"
+      }
+    }
+  }
+}
--- a/apps/launcher/images.json
+++ b/apps/launcher/images.json
@@ -110,6 +110,12 @@
   "covers/refl-vue-sfc-lab-main.png": {
     "linear": true
   },
+  "covers/cover-iphone16-demo-main.png": {
+    "linear": true
+  },
+  "covers/refl-iphone16-demo-main.png": {
+    "linear": true
+  },
   "covers/cover-zoomlab-main.png": {
     "linear": true
   },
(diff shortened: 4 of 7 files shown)


doodlewind:
## Review — content is right, but the branch broke the unit gate; fixed and pushed

**Verdict: the demo itself is clean — after the pushed fix the full gate is green, so this merges once the stack under it does** (it stacks on #257, which has open design findings of its own).

### What I ran

- `bun run test` on this branch as pushed by the PR: **FAIL in the unit stage** — `tests/platform-contracts.test.ts` "every committed demo manifest lands on the expected admission matrix" (378 pass / 1 fail): `apps/iphone16-demo/pocket.json` landed with no matrix row, and that test fails by design for every unlisted manifest. The PR body's validation ran `bun test tests/ios-profile.test.ts` only — the admission matrix and the launcher stages live elsewhere in the gate, which is exactly why `bun run test` is the bar.
- With the pushed fix: **11/11 stages green in 40.5 s**.
- The demo renders end-to-end through the pocket-apple host (`bun tools/build.ts iphone16-demo-main`, then the render_hero example): 180 frames, non-blank, byte-identical across two instances — "Tap Me", the iPhone 16 Pro device line, and the FPS tile all correct. Built at `--hz=120` the headline and tile read 120. (Caveat that belongs to #257, not this PR: the host in that experiment never called `set_tick_rate`, and nothing objected to the mismatch.)

### Pushed to this branch (822dc4c)

1. `"iphone16-demo": [false, true, false]` in the admission matrix — it resolves exactly like nsengine (vita-only through the shared touch + integer-fit contract, private ios-dev otherwise).
2. The launcher display union regenerated (`bun tools/launcher.ts scan`) with both vita-only demos, and the Vita-equality pin in `tests/launcher-sim.test.ts` evolved to subset + exact delta. Note: this overlaps the review commit on `feat/ios-cli` (f5a5975), which makes the same registry/test change for nsengine alone — whichever lands second reconciles trivially in the stack rebase.

### Notes

- `app.tsx`/`main.tsx` follow the repo conventions (framework imports from `@pocketjs/framework/*`, mounted entry, `// @title` header) and the manifest mirrors `iphone2g-demo`'s shape with the 480×272 integer-fit viewport. Nothing else to flag in the 42 added lines.
- One consequence of the manifest worth being deliberate about: like nsengine, this demo now appears in the committed launcher display union and is Vita-admissible by contract, so it shows up in Vita's Cover Flow deck. If iOS demos shouldn't ride along there, the exclusion belongs in the manifest/admission story, not the registry file.

doodlewind:
Status update: **#255 and #256 are merged.** Your `d00b12cb` (`chore: use latest ios-quickjs`) landed inside #256's merge — the pin lives in `hosts/apple/ns-shell/package.json`, which is #256's file, so it was cherry-picked onto `feat/ios-cli` with authorship preserved rather than waiting for this branch. It will fall out of this branch on rebase as already-applied.

This PR is otherwise ready (gate was 11/11 green with the review fix) and now waits only on **#257** — once that merges, rebase onto main and the remaining diff should be just `apps/iphone16-demo/` plus the admission-matrix/launcher rows from the review commit.

doodlewind:
Rebased onto the current #257 head and pushed, so this branch is a coherent stack again instead of conflicting against main.

- `git rebase --onto <257-head> 06cf3b1a` — replays this PR's own two commits onto the rebased tick-rate branch. **`d00b12cb` dropped automatically** ("patch contents already upstream"), exactly as expected: the ios-quickjs pin landed inside #256's merge.
- One conflict, in `tests/launcher-sim.test.ts`: main carries the nsengine-only Vita delta from #256's review fix, this branch carries the two-demo version. Resolved to the union — `["iphone16-demo-main", "nsengine-main"]`, PSP 17 / Vita 19. `bun tools/launcher.ts scan` confirms the committed registry is already fresh against that.

Verification: `bunx tsc --noEmit` clean, `bun run test` **11/11 stages green**.

The branch is now `tick-rate → tick-rate review fix → iphone16-demo → iphone16-demo review fix`. Once #257 squash-merges this needs one more rebase onto main, after which the diff should be just `apps/iphone16-demo/` plus the admission-matrix and launcher rows.

One note on the gate, since it cost me a couple of reruns: the unit stage failed 383/1 twice here, including with no concurrent run, but the same 38 files pass 384/0 standalone and a full-gate rerun goes 11/11 green in 39 s. That is the known macOS Gatekeeper first-launch stall on freshly linked test binaries (`symbian-runtime` / `iphone2g-device-transaction`), which exists on main — not anything this branch introduced.

NathanWalker:
Rebased onto the new #257 head (the three review fixes just pushed there), replaying this PR's two commits — no conflicts, the demo files are disjoint from the tick-rate changes. Verified on the stack: `bun run test` **11/11 green (37.4 s)**, `bunx tsc --noEmit` clean. Once #257 squash-merges, one more rebase and the diff should still be just `apps/iphone16-demo/` plus the admission-matrix and launcher rows.

doodlewind:
Merged as `dcbf536`, right after #257. Rebased onto the post-#257 main (`--onto origin/main d6ae5f03`) so only this PR's two commits replay; net diff is exactly the seven files predicted — `apps/iphone16-demo/` plus the admission-matrix and launcher-union rows. Gate 11/11 green (36.9 s), `bunx tsc --noEmit` clean.

PR #274, opened 2026-08-14
chore(ios): pin ns-shell to @nativescript/pocketjs 0.2.1

NathanWalker:
The one-line pin bump flagged in the #257 thread. `@nativescript/pocketjs@0.2.1` is the plugin rebuild from post-merge main ([NativeScript/pocketjs#1](https://github.com/NativeScript/pocketjs/pull/1), merged): the xcframework is built from upstream `main` after #257/#258 landed (so it also picks up #270's allocator change), `PocketHostView._mountUi` publishes `__tickHz` next to `__host`/`__hostAbi`, and `tickRate` applies ahead of guest eval in both views — completing the pairing invariant for external-guest mounts.

Validated against the rebuilt binary on the iPhone 17 Pro Max simulator (iOS 26.5) with `nsengine --hz=120 --density=4`: guest mode (`[pocket-shell] guest loaded`, pong round trip, "sandboxed realm") and `--external-guest --no-build` (build stamp staged 120 Hz, "iOS 26.5 via NativeScript"). `render_hero` at declared 120 Hz: 180 deterministic non-blank frames, damage_px matching the #257 merge-time verification.

**Draft until `0.2.1` is live on npm** (publish is in flight); merging before then would break the shell's `npm install`.

### DIFF
--- a/hosts/apple/ns-shell/package.json
+++ b/hosts/apple/ns-shell/package.json
@@ -5,7 +5,7 @@
   "main": "src/app.ts",
   "dependencies": {
     "@nativescript/core": "9.1.0-alpha.11",
-    "@nativescript/pocketjs": "0.2.0"
+    "@nativescript/pocketjs": "0.2.1"
   },
   "devDependencies": {
     "@nativescript/ios-quickjs": "9.0.0-preview.3",


NathanWalker:
`@nativescript/pocketjs@0.2.1` is live on npm, so this is out of draft. Validated the pin against the registry, not the local checkout: cleaned `hosts/apple/ns-shell/{node_modules,platforms}`, then `pocket ios play nsengine --no-build` with **no** `--plugin-path`/`--runtime-tgz` overrides — npm resolved 0.2.1, both xcframework slice binaries in the installed package are byte-identical to the merged plugin main, and the 120 Hz bundle boots on the iOS 26.5 simulator (`[pocket-shell] guest loaded`, pong round trip, "sandboxed realm" tile). A non-60 bundle mounting at all is the contract working: 0.2.1's mount publishes `ui.__tickHz`, which 0.2.0 refuses.

NathanWalker:
Just a note @doodlewind with all settled, last thing I'll do is followup after 9.1 is released (targeted before end of this month) to ensure the apple/ns-shell uses latest versions of core and engines.

doodlewind:
Merged. Verified the pin against the published tarball rather than the version string:

- `pocket_apple.h` carries the **post-#257 order** — `create -> load_pak* -> [set_identity] -> [set_tick_rate] -> eval_bundle` — so the header that shipped 0.2.0's too-late ordering is gone.
- `index.js`: `_mountUi` publishes `__tickHz` alongside `__host`/`__hostAbi`, and `createNativeView` sets `tickRate` before `evalBundleLabel`. External-guest mounts now complete the pairing invariant instead of tripping it.
- The arm64 binary differs in size from 0.2.0's — a real rebuild, not a repack.

Gate 11/11 green, `tsc` clean, and nothing in `tests/` pins the version string, so this is the whole change.

This retires the known limitation recorded in the 0.10.0 changelog entry. That entry stays as written — it was accurate at release time — and the next release's entry is the right place to note that the iOS 120 Hz path is now unblocked end to end.

PR #275, opened 2026-08-14
fix(audio): audioFramesForTick counts frames at the realm's declared rate

NathanWalker:
The audio deferral from #257, closed before any audio-capable host grows a declared rate: `audioFramesForTick` divided by a literal 60, so a 120 realm would consume PCM at 2x and the zero-drift statement went false.

### What changed

- **`contracts/spec/audio.ts`**: `audioFramesForTick(rate, tick, ticksPerSecond = 60)`. The default keeps every existing callsite byte-identical — the 60-window tests now pin the default path. The doc states the generalized property (sum over any `ticksPerSecond` consecutive ticks is exactly `rate`) and the reachability argument: the pairing invariant refuses a mismatched mount, so formula and realm can only disagree through a host bug, never through a bundle.
- **`hosts/sim/audio.ts`**: `createSimAudioSink(ticksPerSecond = 60)` — a runner driving a non-60 realm constructs the sink at that rate, keeping one-tick-one-`sink.tick()` pairing intact.
- **`contracts/spec/gen-rust.ts` + regenerated `engine/core/src/spec.rs`**: the pin now also asserts the formula at a declared 120 (`22050 -> 183, 184`), and the emitted doc states the rate-parameterized form.

### Tests

- Zero drift at declared 90/120/240 for every `AUDIO_RATES` entry, window starts 0/1/59/600, plus the ±1-frame bound.
- Sink x WavPlayer at a declared 120: the same 1 s track consumes exactly 22050 frames in 120 ticks with zero underruns, and the consumed byte-stream hash equals the 60-sink run — the declared rate changes per-tick chunking, never the bytes.

### Verification

`bun run test` **11/11 stages green (38.5 s)**; `bunx tsc --noEmit` clean; `bun tests/contract.ts` green against the regenerated spec.rs; `cargo check -p pocketjs-core` clean (the spec.rs delta is doc text).

Deliberately not here: nothing calls this with a non-60 rate yet — the sim can't reach a non-60 realm until `engine/wasm` exports `ui_set_tick_rate` (next PR in this follow-up series).

### DIFF
--- a/contracts/spec/audio.ts
+++ b/contracts/spec/audio.ts
@@ -136,18 +136,28 @@ export const AUDIO_MAX_STREAMS = 4;
 
 /**
  * Frames the audio clock consumes on virtual tick `tick` (0-based, counted
- * per stream while playing) at 60 core ticks per second. The Bresenham-style
- * floor difference distributes non-divisible rates (22050/60 = 367.5) with
- * zero drift: the sum over any 60 consecutive ticks is exactly `rate`.
+ * per stream while playing) at `ticksPerSecond` core ticks per second — the
+ * realm's declared rate, 60 unless the host declares otherwise. The
+ * Bresenham-style floor difference distributes non-divisible rates
+ * (22050/60 = 367.5) with zero drift: the sum over any `ticksPerSecond`
+ * consecutive ticks is exactly `rate`.
  *
  * This formula IS the deterministic-audio contract: virtual-clock hosts
  * (hosts/sim) consume exactly this many frames per playing tick, making the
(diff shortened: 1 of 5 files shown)


NathanWalker:
A better way to describe this:
> The deterministic audio sink converts source sample rate into frames-per-core-tick using the realm's fixed-step tick frequency.

Instead of `audioFramesForTick(rate, tick, ticksPerSecond)`, probably better as:
```
audioFramesForTick(
  sampleRate: number,
  tick: number,
  tickHz = 60
)
```

PR #283, opened 2026-08-15
fix(tools): build desktop launchers from the engine workspace

jnhu76:
## What

Fix the `note` and `widget` launchers after the repository layout
migration in #149.

#149 moved the desktop Cargo workspace from `pocket3d/` to `engine/`,
but these two launchers retained the old build cwd and an incorrect
nested target path.

Both launchers now:

- build from the `engine/` workspace
- resolve binaries from `engine/target/release/`

## Scope

This only fixes the stale Cargo workspace paths.

Windows filesystem URL handling and Windows desktop support are
intentionally out of scope.

## Verification

From the `engine/` workspace:

- `cargo build --release -p note-widget`
- `cargo build --release -p pocket-stage`

Both packages resolve from the workspace and emit their binaries under
`engine/target/release/`.


doodlewind (approved):
Thanks for the fix!

PR #314, opened 2026-08-21
feat(site): a Beyond 2D UI chapter, and plainer landing copy

doodlewind:
Follow-up to #311 and #313.

## Beyond 2D UI

A new chapter between Architecture and Ecosystem, for the thing an application runtime is least expected to do. PocketJS is an application runtime with a game engine's architecture, so a game and an app are built the same way.

The visualization holds one guest program above a row of native cores, and a single 10.5s cycle rotates three compositions:

| composition | cores mounted |
|---|---|
| Pocket Talk | `ui` + `net` |
| OpenStrike | `ui` + `audio` + `strike` |
| Pocket Voxel | `ui` + `audio` + `voxel` |

The composition name, the lit cores and the screenshot all switch together, so the three read as the same machine being reassembled rather than three separate diagrams. The name and the screen switch on a **hard cut**: a crossfade overlapped two names into garbled text and two screenshots into a double exposure.

Cores are described the way they behave: independent native modules, loaded the way a kernel loads drivers, and absent from the build when the content does not ask for them.

## Copy

Several passages said what they were doing instead of what they meant, so the section ledes are now statements: the Modern DX lede names QuickJS in one sentence, the Pocket Lab band is one sentence, and the ecosystem lede stops repeating the module vocabulary that the new chapter now explains.

## Two fixes

- `docs/architecture` and `docs/frameworks` in the Modern DX refs were pointing under `/docs/overview/`, which does not exist. A pass over every internal link on the homepage confirms the rest resolve into `site/dist`.
- Panel notes were using `class="foot"`, which is the site footer's class, so they inherited its padding and background. They are `pnote` now, which also fixes the same problem in the Architecture and Modern DX panels.

A 3D version of the pipeline comparison was tried and reverted; the flat one reads better.


doodlewind:
Superseded by a branch cut from current main; this one carried the already-squashed commit from #313 and conflicted.

PR #318, opened 2026-08-22
fix(site): give iOS every apple-touch-icon rung it asks for

doodlewind:
Adding pocketjs.dev to an iPhone home screen still produced a page screenshot instead of the mark, after #317 shipped the icon family.

## What was not wrong

The file and the way it is served both check out, so I ruled these out before changing anything:

- `/apple-touch-icon.png` returns **200 `image/png`**, and the bytes from the edge are **byte-identical** to the file in the repo (same SHA-256)
- the PNG is **180×180, 8-bit, colour type 2 (RGB, no alpha), non-interlaced** — nothing exotic for iOS to choke on
- the `<link rel="apple-touch-icon">` is present in the HTML **as an iPhone Safari UA receives it**, on every page
- per WebKit, since Safari 15.4 `apple-touch-icon` **takes precedence over** the manifest's `icons`, so the manifest was not stealing the choice

## What was wrong

iOS picks the `apple-touch-icon` whose `sizes` is closest to the size it wants, and a single 180×180 link is a weak answer to that: an iPad asking for 152 or 167 has nothing to match, and a client reading `sizes` strictly has no other candidate. On top of that, **iOS caches the home screen icon per URL** and keeps serving a page screenshot once a fetch has come up empty — which is what a device that visited the site before #317 would have recorded.

So:

- the ladder is declared in full — **180 / 167 / 152 / 120**, each rendered at that exact size rather than scaled from one file
- `apple-touch-icon-precomposed.png` is at the root for clients that read no link tag and just fetch that path
- every URL carries **`?v=2`**, defeating the per-URL icon cache; `ICON_V` is a constant in `templates.ts` to bump when the drawing changes
- the manifest no longer lists `favicon.svg`: no home screen on either platform renders an SVG, and its `"sizes":"any"` outranks the rasters when a picker sorts by size
- `apple-mobile-web-app-title` keeps the home screen label at "PocketJS" instead of the full tab title

## Verification

- `bun test tests/site-stage.test.ts` — 13 pass. A new assertion ties markup to bytes: the rungs in `ICON_LINKS` must be exactly 180/167/152/120, and each named PNG's IHDR must carry the size its link claims.
- built output: all four rungs plus the precomposed file land at the site root; all four pages (home, docs, playground, /for/) carry the four links.
- 120×120 render inspected directly, since that is the size where the 11% inset could crowd the mark.

Same change is going up on pocketlab.build (pocket-stack/pocketlab.build#6).

Note for the device: iOS will not re-fetch for a home screen entry it already made. Delete the old one, reload pocketjs.dev in Safari, then add it again.


doodlewind:
Superseded by a branch cut from main: this one was pushed onto the already-squash-merged #317 branch, so its diff replayed merged work. The identical change is in #319.

PR #330, opened 2026-08-23
fix(compiler): restore Windows path portability

Chronicle-IRAS:
## Summary

Restore Windows-safe compiler path handling and add regression coverage
for compiler portability.

## Problem

PocketJS compiler code paths were mixing URL pathname semantics with
filesystem path consumers.

On Windows, URL.pathname can produce paths like:

/C:/...

which are not valid native filesystem paths for Bun filesystem APIs.

Compiler path matching also assumed `/` separators, and JSXText comparison
did not account for CRLF normalization performed by Babel.

## Root cause

Filesystem paths, URL paths, and module resolution paths were treated as
the same representation.

Some portability handling previously introduced for Windows was lost during
later compiler changes, allowing these assumptions to return.

## Changes

- Use fileURLToPath for compiler-owned filesystem paths.
- Keep compiler source hashing in URL pathname space where required.
- Normalize Windows separators for compiler path matching.
- Fix generated-style path detection on Windows.
- Normalize JSXText CRLF comparison without changing output text or
  allowing HTML entities.
- Add compiler portability regression tests.

## Historical context

Related:

- #122
- #193

#122 introduced broader Windows portability fixes.

#193 later addressed overlapping compiler file URL handling but was closed
without merging.

This PR revisits the issue against current main and additionally covers
current compiler registry paths, separator handling, generated-style build
walking, CRLF behavior, and regression tests.

The public record does not state why #193 was closed.

## Validation

Validated on Windows:

- git diff --check
- compiler portability regression tests
- bun run test:tailwind
- bun run dev
- Browser host HTTP validation
- Solid build
- Vue Vapor build
- Octane build

Regression coverage includes:

- filesystem path consumers
- cache/runtime paths
- Windows separator handling
- LF/CRLF JSXText behavior
- entity rejection behavior

## Known limitations

- Repository currently has no Windows CI runner.
- Some existing Windows baseline tests outside this patch still fail due
  to unrelated pathname assumptions and missing platform tools.
- PSP/Vita/Playdate/device toolchains were not tested.

### DIFF
--- a/framework/compiler/jsx-plugin.ts
+++ b/framework/compiler/jsx-plugin.ts
@@ -6,6 +6,7 @@ import solidPreset from "babel-preset-solid";
 import tsPreset from "@babel/preset-typescript"; // untyped - see framework/compiler/ambient.d.ts
 import { transformVueJsxVapor } from "vue-jsx-vapor/api";
 import { existsSync } from "node:fs";
+import { fileURLToPath } from "node:url";
 import { compileVueSfc } from "./vue-sfc-compile.ts";
 import {
   propsHelperCode,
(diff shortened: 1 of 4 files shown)


doodlewind:
Thanks for restoring Windows-safe compiler path handling and adding focused regression coverage for native paths, separator handling, and CRLF behavior. The implementation keeps the URL/filesystem boundary explicit, and the validation is green. Merged—thank you for the contribution!

PR #335, opened 2026-08-26
feat(esp-idf): add composable PocketJS components

HalfSweet:
## Summary

- Provide six caller-composed ESP-IDF components and an optional runner. Product firmware owns tasks, input, buffers, presentation, and package storage.
- Support project-owned host profiles and fixed-width package admission without NET/FS/DB adapters.
- Keep this PR draft. ESP-IDF validation remains manually dispatched, not triggered on every PR or push.

## Review fixes

- Expose borrowed PAK bytes with `JS_SetImmutableArrayBuffer`, without a full RAM copy. The pinned QuickJS 0.14.0 source is prepared in the build directory with missing reverse/species-write checks; its source hash is verified first.
- Stage all PAK assets and registration memory before committing. Returned errors preserve core/binding state and permit retry. Also correct the C PAK directory stride from 20 to 24 bytes.
- Invalidate borrowed frames on every draw, mutation, and tick. Prepared targets track both core identity and generation.
- Reject raster densities outside 1..=255 at the FFI boundary. Document fatal Rust OOM instead of promising universal recoverable allocation.
- Replace configure-time application globs with compiler depfiles covering the actual bundler graph, resources, configuration, and tooling inputs. Support additional compiler receipts and custom dependencies.
- Split native core and renderer archives and vendor trees. Move the general backend to `engine/backends/rgb565`; renderer resource access uses the C ABI, not private Rust `Ui` layout. Isolate non-standard-library Rust symbols per component while retaining the shared standard-library runtime needed by S3.
- Verify native source/archive digests, sizes, build policy, compiler pins, and preparation-tool identity before staging. Archive preparation and compiler-wrapper rules participate in source identity. Component versions and dependency ranges have a checked release map.
- Replace public `idfHost` fields with a versioned, hashed `hostExtension` payload interpreted by the IDF adapter.
- Capture UI bindings in native closures instead of context opaque. A guest-owned name registry rejects a second UI mount.
- Generate C/Rust ABI data declarations, function prototypes, layout assertions, and wire constants. Share golden/malformed package cases across TS, Rust, C, and Python; add a C fuzz target.
- Track Promise identity, release reported records, and tolerate reentrant rejection reporting. Propagate JS conversion exceptions and set proper OOM exceptions.
- Update integration and maintainer documentation, and refresh the deterministic IDF package fixture.

## Validation on the review-fix revision

- Canonical repository suite: **11/11 stages passed**. Fixture Git signing was disabled only for the test process; no Git configuration was changed.
- Rust core and RGB565 backend tests passed, including the shared malformed corpus and asset transaction regression.
- Real C/QuickJS/Rust host regressions passed: immutable writes, failed PAK retry, C allocation fault injection, Promise identity/reentrancy, context opaque coexistence, second mount rejection, and stale frames.
- Real CMake/Ninja incremental test passed: new import, subsequent edit to that import, image-only edit, and compiler-receipt-only edit all rebuild; unchanged inputs do not.
- ESP-IDF 6.0 P4 and S3 firmware builds passed with both prebuilt and Rust-from-source archives.
- Both targets' archives have no overlapping private `pocketjs_core` symbols. All four build receipts passed verification.
- Seven-component release staging passed. Both vendored native source trees compiled offline for P4.
- Website build passed: 19 documentation pages.
- C parser fuzzing: **10,000 libFuzzer + UBSan runs passed**. Full ASan/UBSan is configured in the manual Linux workflow. Local Homebrew ASan stalled in runtime initialization before reaching the harness, so it is not counted as a passed test.

## Remaining release gates

No hardware was flashed in this round. P4/S3 pixel hashes, hardware PPA counters, long-running heap/cadence/stop checks, and the complete IDF 6.0/6.1 release matrix remain release validation work. No Registry upload or merge was performed.


doodlewind (approved):
LGTM if you're moving your esp-idf repo here :)

PR #345, opened 2026-08-30
feat(3ds): support companion transport alongside offload

doodlewind:
Pocket Term needs the 3DS host to implement `svcOpen`, `svcPoll` and `svcSend`; a build pinned to main previously compiled but could not connect to its Mac terminal daemon. This adds the PKNT companion transport, runtime font uploads and the desktop mirror's remote companion adapter.

The 3DS transport discovers the matching app over UDP, uses nonblocking TCP with bounded frame pumping, recovers after disconnects, and preserves the connection while clearing guest-visible input on a guest reload. Capture builds keep the transport inert. Node companions can import the shared font baker through opentype.js's CommonJS default export.

This branch is integrated with the current main, including Pocket Doc's offload/resources and the per-app runtime storage from #355. DevTools, companion sockets and the offload worker share one process-wide SOC instance. An atomic state prevents concurrent initialization without making a competing UI caller wait; failures retry after a cooldown. Offload initialization and retries remain on its worker. Shutdown joins that worker and stops all transports before releasing SOC. Existing guarded New 3DS input initialization and the 384 KiB QuickJS stack are retained.

Validation:
- 35 3DS profile/storage/wire/SOC tests passed. The new native C harness exercises simultaneous initialization, immediate rejection of the competing caller, retry cooldown, allocation failure, recovery and shutdown under ASan/UBSan.
- 35 resource/offload tests passed; generated contract checks passed.
- `cargo check --locked --manifest-path hosts/desktop/Cargo.toml` passed.
- Full ARM `.3dsx` builds of Pocket Shell and Pocket Term passed with the combined runtime and separate app slots.
- Added a path-filtered 3DS runtime CI job for the native harness and existing profile/storage/wire checks.

Shell and Term will pin the merged runtime and receive the current 24 px/48 px HBL icon assets. Fresh physical interaction with those rebuilt apps is separate from the already confirmed Pocket Doc hardware run.


doodlewind:
**Hardware-verified.** After reflashing the rebuilt `.3dsx` (new Homebrew entry; the contacts-demo runtime left on card as fallback), the real console beacon-discovered the Mac daemon and ran live: zsh prompt + colored output over WiFi, three concurrent sessions with tab switching, the touch keyboard's one-shot Ctrl highlight, and the red non-zero-exit prompt — all driven from the device.

Two deployment lessons landed as follow-ups in the branch:
- `bun run e2e:3ds` overwrites `dist/3ds/<output>.3dsx` with the **capture** build (svc stubbed, parks after its tape) — the first two reflash attempts shipped that binary and looked like a frozen connect screen. Rebuild with `bun tools/3ds.ts <app>` after any golden run before flashing; `strings <3dsx> | grep -c pocketjs-captures` must be 0.
- SOC/devserver init no longer latch a boot-time failure: soc_ensure retries behind a 3 s cooldown and the main loop re-runs a failed devserver_init every ~5 s.

doodlewind:
## Round 2: mirror windows, runtime CJK, closable tabs, held ZL

**A read-only desktop window per session.** `apps/term-mirror` is the console's top screen and nothing else — the same grid component (`apps/term/grid.tsx`, extracted for this), the same store, the same protocol, drawn by PocketJS through gpui instead of the PICA200. The companion opens a window when a session starts and closes it when the session ends. `role: "mirror"` makes the daemon refuse input from that connection and never resize its PTY, so watching cannot disturb the console. Because `hosts/desktop` is the stock host of **macos-app and linux-app**, the window follows the console to Linux with no new work.

That needed the desktop host to speak the wire: **`hosts/desktop/src/net.rs`** adds a PKNT client behind `--svc-connect host:port`, reusing `pocketjs_core::wire` (already linked) and threaded like `hosts/vita/src/net.rs` so the fixed-step tick never blocks. When a wire is present it owns the svc queues and the in-process editor dialect sees an empty drain.

**CJK without a font on the device.** The console bakes atlases from the app's own string literals, so a session printing 你好 had nothing to draw with. The companion now rasterizes what is actually on screen — the repo's own `bakeSlot` over a system face, lifted out of a `.ttc` because every CJK font macOS ships is a collection — into a FONT ATLAS for spare slot 19, streamed as paced base64 chunks the guest loads through the spec `loadFontAtlas` op. It is the note widget's runtime-coverage design (docs/WIDGET.md) with the rasterizer moved to the companion. Advances are rewritten to whole cells so mixed CJK/ASCII stays on the grid, and xterm's zero-width continuation columns are skipped rather than spaced.

**Closable tabs** (× on the active tab) and **ZL as a held Ctrl**, with the on-screen keyboard's Ctrl cap lit from the same state. ZL needed native work: it is New-3DS-only and arrives through **ir:rst**, not the HID pad, so `input.c` brings that service up, and `BTN.ZL`/`BTN.ZR` take two bits the PSP never assigned. An Old 3DS never sets them and L held is the same modifier there.

### Verified
- Console and mirror window showing the same session and the same CJK line at the same moment; two sessions opening two windows; the daemon dying and the console reconnecting to a fresh one on its own in ~5 s.
- One real bug found by testing: the native transport outlives the guest, so after a hot push the daemon still thought the atlas had been delivered and CJK rendered as blanks. A `hello` now means "this replica has loaded nothing", not "new socket".
- 12/12 test stages, 18/18 3DS goldens byte-exact, contract drift guard green (the BTN addition regenerates both `spec.rs` and the iphone2g C header).

### Needs a reflash
ZL is native. `dist/3ds/pocketterm-main.3dsx` is rebuilt and carries it; everything else in this round is hot-pushable and already running on the console.

PR #352, opened 2026-09-01
fix(3ds): the runtime fixes Pocket Shell found — adjacent-swap reconcile, the guest's stack budget, a pinned emulator clock

doodlewind:
## What

Pocket Shell now lives in its own repository — **[pocket-stack/pocket-shell](https://github.com/pocket-stack/pocket-shell)**, with this checkout as its `vendor/pocketjs` submodule, the same split Pocket Term took. This branch keeps only what belongs to the runtime.

Three fixes, each found by running a real UI on the console rather than a demo:

- **`framework/src/native-tree.ts` — `insertBefore(node, node)` is a no-op, like the DOM.** Solid's `reconcileArrays` issues exactly that call when two *adjacent* `<For>` items swap; the old code unlinked the node first, so it became its own missing anchor and threw `insert anchor is not a child of parent`. Every adjacent swap crashed. Any app that reorders a `<For>` by one position hits this; `tests/renderer.test.ts` pins the case.
- **`hosts/3ds/src/qjs.c` — the guest's JS stack is 384 KiB, not 192.** The console spends it on JSX nesting *depth*, not node count: a QuickJS call frame is expensive and mounting descends the tree. A UI four levels deeper than a demo overflowed it mid-frame, after which the runtime's recovery chain fell back to last-good — which on a shared SD card looks exactly like a different app booting. This is the same budget the Pocket Term work independently needed.
- **`tests/e2e/azahar.ts` — the fixture pins the emulated clock** (`init_clock=1`, `init_time` = 2000-01-01), so a guest that shows the RTC can still be compared byte for byte. `DEFAULT_SPEC_NAMES` is back to the demo.

## Verified

- `bunx tsc --noEmit` clean; `tests/renderer.test.ts` 51/51 and `tests/platform-contracts.test.ts` 37/37 green after the removal.
- The extracted app, built out-of-tree against this commit, **reproduces all 26 pinned frames byte for byte** against the recordings made while it still lived here — so the move changed nothing about what the console draws.

The app, its tests, its input tapes and those frames now live in the product repository, along with a frame recorder that turns the same tapes into the animations its README shows.


doodlewind:
Closing as superseded by current main. Main already contains the 384 KiB QuickJS stack budget, the deterministic Azahar RTC settings, and the adjacent-swap renderer regression test. Its insertNode implementation also validates the anchor before treating self-insertion as a no-op, so merging the older early-return guard here would duplicate and bypass that validation. Pocket Shell is being updated to the consolidated main runtime.

PR #356, opened 2026-09-03
perf(core): cut peak live memory 17.3% (taffy capacity, MeasureCtx dedupe)

quake:
## Summary

Cuts `pocketjs-core` peak live memory **1,429,840 -> 1,182,457 bytes (-17.3%)** on the new deterministic `membench` journey, with zero behavior change (drawlist checksum byte-identical, 127 tests pass, wasm32 no_std build clean).

- Drop the unread `MeasureCtx.text` duplicate of every text run
- Rebuild the taffy tree at subtree-sized capacity, freeing the old maps first (slotmap geometric growth left ~160 dead slots x ~1 KB)
- Slim `MeasureCtx` to the shaped size; drop per-level children clones in `collect_subtree`/`collect_run`

`engine/core/examples/membench.rs` is the new counting-allocator benchmark (boot -> steady -> churn -> burst; prints `PEAK_LIVE_BYTES`).

## Allocation-churn trade-off (measured)

Structural rebuilds now re-allocate the taffy maps instead of reusing historical capacity:

| | baseline | head |
| --- | --- | --- |
| `TOTAL_ALLOC_BYTES` | 11,416,332 | 14,755,184 |
| `ALLOC_COUNT` | 109,081 | **101,314 (−7.1%)** |

The +3.3 MB of total allocation bytes is a handful of large blocks per structural rebuild (recycled by the PSP arena's O(1) size-class free lists); the allocation **count** goes down.

## Benchmark integrity

`NODES` reports the true live node count via a harness-side hierarchy model mirroring `Ui::destroy_node`'s subtree semantics (a call-counting counter drifted by +35 after churn). `DRAWLIST_CHECKSUM` is byte-identical across baseline and head.

## Note on CI

Workflow runs on this fork PR require maintainer approval (`action_required`); the numbers above are local runs of `cargo test`, `cargo run --example membench`, and `cargo build --target wasm32-unknown-unknown`, reproducible with the commands in the diff.

### DIFF
--- a/engine/core/src/layout.rs
+++ b/engine/core/src/layout.rs
@@ -23,12 +23,12 @@ use crate::text::Fonts;
 use crate::tree::{LayoutRect, Tree};
 
 /// Measure context attached to text leaves (taffy NodeContext).
+///
+/// Deliberately minimal: the solve-time measure closure reads only `size`,
+/// and shaping happens once in `MeasureCtx::shaped` (build/restyle) against
+/// the freshly collected run — storing slot/tracking/line_height here (or
+/// the run itself) would be dead weight duplicated per text node.
 pub struct MeasureCtx {
-    pub text: String,
-    pub slot: u8,
-    pub tracking: f32,
-    /// NAN = atlas default.
-    pub line_height: f32,
     /// Shaped size, computed ONCE when the context is (re)built. Text
     /// shaping is the expensive half of layout on the PSP; the taffy
     /// measure closure must never re-shape per solve pass.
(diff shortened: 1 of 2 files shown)


HalfSweet:
Thanks for the optimization work. Removing the duplicated data from `MeasureCtx` and avoiding `children.clone()` during recursive traversal both look reasonable.

My main concern is that the current benchmark measures live bytes using the host allocator, while the actual PSP constraint is the arena high-water mark. The benchmark also includes some allocations from the test harness itself, so the current results are not enough to show that rebuilding and resizing the Taffy tree on every structural relayout is a net improvement on PSP.

There are also a few unrelated or incomplete benchmark artifacts in this PR, along with some inconsistencies between the benchmark code and its comments. I suggest splitting the work: merge the low-risk deduplication and clone removal first, then fix the benchmark and provide PPSSPP/PSP `arena_bump_bytes` and layout timing results before evaluating the Taffy storage rebuild strategy separately.

For now, I’m requesting changes.

quake:
I have narrowed this PR to the two low-risk core changes:

- keep the `MeasureCtx` data deduplication
- keep the recursive traversal changes that avoid `children.clone()`
- remove the Taffy capacity-rebuild strategy
- remove the unrelated/incomplete benchmark artifacts

The Taffy storage rebuild will be proposed in a separate PR after adding PSP/PPSSPP `arena_bump_bytes` and layout timing measurements. This PR now passes the core test suite: 127 tests passed.

HalfSweet (approved):

PR #361, opened 2026-09-04
feat(quickjs-c): add external harness hooks

HalfSweet:
## Summary

External benchmarks need to attribute guest evaluation, frame execution, pending jobs, and core ticks while driving repeatable actions against the same retained UI instance. This change adds opt-in hooks to `engine/quickjs-c` and `engine/ui-cabi`; the external harness owns action protocols, measurement, and verification.

- `POCKET_RUNTIME_STAGE_HOOKS` reports stage boundaries through `pocket_bench_stage`.
- The separate `POCKET_RUNTIME_HARNESS` capability caches one guest dispatcher and invokes it with two integers through `JS_Call`, without evaluating source text, changing stages, or draining jobs.
- `harness-access` exposes an unsafe closure over the initialized C ABI `Ui`, with a no-concurrency/no-reentrancy lifetime contract.
- Production hosts leave these opt-ins disabled.

Two unconditional correctness fixes support the same workloads: validate foreign insertion anchors before either tree changes, treat insertion before the same node as a positional no-op, and accept the 16-byte texture allocation alignment required by supported 64-bit C hosts.

## Runtime contract

```text
boot success:       EVAL, IDLE, JOBS, IDLE
boot eval failure:  EVAL, IDLE
boot frame missing: EVAL, IDLE
boot job failure:   EVAL, IDLE, JOBS, IDLE
frame success:      JS, JOBS, TICK, IDLE
frame JS failure:   JS, IDLE
frame job failure:  JS, JOBS, IDLE
```

The embedding sets the stage for synchronous dispatcher work. Queued jobs run during the next guest turn's `JOBS` interval. The accessor returns `None` when the singleton is absent and does not create a replacement instance. See `engine/quickjs-c/README.md` for the integration and validation contract.

## Regression coverage

The `Native C harness` workflow runs on pull requests and main changes on Linux and macOS:

- Compile, link, and execute the real `pocket_runtime.c` against deterministic QuickJS/UI stubs in all four stage/dispatcher configurations.
- Run renderer, virtual-list, and Vue DOM contracts.
- Test `ui-cabi` with `harness-access` enabled.
- Build the real `bare-platform,software-only` static library with its pinned nightly, link a C executable, and upload 200 textures per allocator variant through both C `malloc` and host-owned callbacks. Exercise texture-table growth, explicit frees, shutdown, and reinitialization; host callbacks verify alignment, allocator calls, and cleanup.
- Check generated contracts for drift.

The Rust texture unit test uses the default allocator; the linked C fixture is the evidence for the `CAllocator` path. QuickJS stubs establish runtime control flow, not real-engine benchmark parity.

## Validation

Local macOS arm64 validation:

- Runtime/renderer/virtual-list/Vue DOM: 73 tests passed.
- `cargo test --locked --manifest-path engine/ui-cabi/Cargo.toml --features harness-access`: 10 tests passed.
- `bun test tests/ui-cabi-allocator.test.ts`: both real allocator variants passed.
- Negative control: the same linked allocator test aborts on the pre-PR base with the old 8-byte alignment limit.
- `bun tests/contract.ts`, workflow actionlint, and `git diff --check`: passed.
- `bunx tsc --noEmit`: existing TS2353 in `framework/compiler/build-inputs.ts:34` (`BuildConfig.write`), reproduced on base `2d20ddad` with the same dependencies.

Contributor-reported downstream validation (not rerun as part of this CI repair): all three host shells build; 70 benchmark unit tests pass; Solid, Vue Vapor, and Octane match their WASM oracle; incorrect action lists are rejected; full/native/raster share RGBA8 hash `cd14be5e`; guest-tape matches 63,154 operations; native replay has zero return-value mismatches. Benchmark adapters retain the action protocol outside PocketJS, and macro applications receive no dispatcher.

### DIFF
--- a/.github/workflows/native-c-harness.yml
+++ b/.github/workflows/native-c-harness.yml
@@ -1,0 +1,78 @@
+name: Native C harness
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/native-c-harness.yml"
+      - "engine/quickjs-c/**"
+      - "engine/ui-cabi/**"
+      - "engine/core/**"
+      - "framework/src/**"
+      - "framework/compiler/**"
+      - "contracts/**"
+      - "tests/fixtures/quickjs-c-harness/**"
+      - "tests/fixtures/ui-cabi-allocator/**"
+      - "tests/quickjs-c-harness.test.ts"
+      - "tests/ui-cabi-allocator.test.ts"
+      - "tests/renderer.test.ts"
+      - "tests/virtual-list.test.ts"
+      - "tests/vue-vapor-dom.test.ts"
+      - "tools/build.ts"
+      - "package.json"
+      - "bun.lock"
+  push:
+    branches: [main]
+    paths:
+      - ".github/workflows/native-c-harness.yml"
+      - "engine/quickjs-c/**"
+      - "engine/ui-cabi/**"
+      - "engine/core/**"
+      - "framework/src/**"
+      - "framework/compiler/**"
+      - "contracts/**"
+      - "tests/fixtures/quickjs-c-harness/**"
+      - "tests/fixtures/ui-cabi-allocator/**"
+      - "tests/quickjs-c-harness.test.ts"
+      - "tests/ui-cabi-allocator.test.ts"
+      - "tests/renderer.test.ts"
+      - "tests/virtual-list.test.ts"
+      - "tests/vue-vapor-dom.test.ts"
+      - "tools/build.ts"
+      - "package.json"
+      - "bun.lock"
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+
+jobs:
+  harness:
+    name: C runtime + allocator (${{ matrix.os }})
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-24.04, macos-latest]
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v7
+      - uses: oven-sh/setup-bun@v2
+      - uses: dtolnay/rust-toolchain@stable
+      - run: bun install --frozen-lockfile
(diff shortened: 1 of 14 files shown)


HalfSweet:
Addressed the review on top of current `main` and force-updated this still-draft branch with `--force-with-lease`.

1. **Boot attribution:** fixed. `JS_Eval` now closes into `IDLE`; boot job draining is bracketed by `JOBS -> IDLE`. The executable harness asserts all four boot sequences and all three frame sequences from the review.
2. **Rust soundness:** fixed. The safe auto-initializing `with_ui` is gone. The replacement is `unsafe with_initialized_ui_unchecked(...) -> Option<R>` with explicit no-reentrancy/lifecycle requirements and before-init/after-shutdown coverage.
3. **Control vs measurement:** clarified and separated. Stage hooks and guest eval now use independent compile-time capabilities. Eval never changes stages; caller-owned attribution and queued-job behavior are documented in both the header and README. In the downstream driver, `run(action)` is explicitly entered under `JS`; the integration run confirmed a non-zero JS phase for synchronous list reversal.
4. **Test quality:** replaced the source-token test with a compiled and linked C harness plus `nm` checks. It injects boot/frame eval and job failures and checks the callback sequences at runtime.
5. **Scope/history:** rewritten as three conventional commits. The PR now states explicitly that only the harness commit is opt-in; renderer and allocator are unconditional correctness fixes.
6. **Anchor validation:** moved before both mirror and host mutation, with a behavior test proving the node remains attached and no host operation is emitted for a foreign anchor.
7. **Allocator contract:** restored the ABI rationale, documented the callback alignment requirement, and added a real texture upload exercising the `Vec<u128>` path. The exact `bare-platform + host-allocator` combination is additionally exercised by the downstream shell build/run.

Full upstream unit + wasm-unit run is green (532 + 216 tests), as are the Rust feature test and the migrated downstream integration described in the updated PR body.

PR #392, opened 2026-09-08
perf(vapor): store the GBA font as 1bpp and expand it at upload

lfkdsk:
## What

The GBA font table `vp_font_tiles` was emitted as 4bpp tiles (3040 B), which was 32.5% of the Todo ROM and 47.4% of the six-button ROM — larger than every code symbol combined. This change emits the font as 1bpp (760 B) and expands it into the 4bpp charblock inside `upload_font` at boot.

## Measured

| ROM | before | after | delta |
|---|---|---|---|
| `todo.gba` | 9356 B | 7160 B | −2196 B (−23.5%) |
| `playdate-six-button.gba` | 6420 B | 4224 B | −2196 B (−34.2%) |

Stable over 5 builds per side. The delta reconciles in the ELF symbol table: −2280 B of table, +84 B of expansion code. State RAM, overlay and reactive figures are unchanged.

## Correctness

- All 3040 bytes that reach the GBA font charblock are byte-identical to the old table, verified by booting the ROM in headless libmgba and dumping charblock 0 (SHA-256 `955a983b…`), not by re-implementing the expansion.
- The existing GBA parity rig cannot see glyph pixels (it probes the screenblock at `0x6004000`; glyphs live at `0x6000020`), so this adds `vapor/tests/gba-font.test.ts`, which reads VRAM back from a booted ROM. Three runtime mutations (halfword order, ink/paper swap, row overrun) each make it fail while parity stays green.
- GB, NES, ESP32 and Playdate generated C is byte-identical; GB/NES final ROMs are byte-identical. ESP32/Playdate final packages were not built locally (no SDK) and their embedded build IDs hash the changed shared `vapor.h`, so only generated-C identity is claimed for those two.
- `bun test vapor/tests/`: 74 pass / 0 fail (was 71), including GBA, GB, NES and oracle parity.

## Provenance

Produced by an autonomous fleet run: a Scout task measured the size baseline and proposed this candidate with a hand-patched prototype; a Builder implemented it; three independent cross-family reviews (different model families) rebuilt the ROMs, re-dumped VRAM and re-ran the mutations. The Scout report lives on the fork branch `lfkdsk/pocketjs:fleet/scout-artifact-size`; this PR carries only the implementation and its tests.

### DIFF
--- a/vapor/compiler/compile.ts
+++ b/vapor/compiler/compile.ts
@@ -2385,8 +2385,9 @@ class AppCompiler {
     const romStrings =
       [...this.strLits.keys()].reduce((a, s) => a + s.length + 1, 0) + this.title.length + 1;
     const pairCount = this.styleTable.pairs.length;
-    const fontBytes =
-      this.target.name === "esp32" || this.target.name === "playdate" ? 95 * 8 : 95 * 32;
+    /* GBA/ESP32/Playdate ship the 1bpp bitmap and expand at boot; GB and NES
+     * bake the two glyph styles into the ROM/CHR table. */
+    const fontBytes = this.target.name === "gb" || this.target.name === "nes" ? 95 * 32 : 95 * 8;
     const styleBytes =
       this.target.name === "gba"
         ? pairCount * 16 * 2 + pairCount + 3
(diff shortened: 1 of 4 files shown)


lfkdsk:

PR #393, opened 2026-09-08
perf(vapor): ship the GB font as 1bpp and expand it at upload

lfkdsk:
## What

Same treatment as the GBA font, for the Game Boy target: emit `vp_font_tiles` as 1bpp and expand it into tile RAM at upload instead of shipping the expanded table in ROM.

## Measured

| | before | after | delta |
|---|---|---|---|
| `todo.gb` occupancy | 11383 B | 9140 B | −2243 B (−19.7%) |
| sdcc `_CODE` | 9335 B | 7092 B | −2243 B |
| `_DATA` | 1143 B | 1143 B | 0 |

5 builds per side. The saving is entirely ROM `_CODE`: the table shrinks 3040 → 760 B, the expansion loop costs 37 B, and no state RAM is traded.

## Correctness

- The 3040 bytes reaching GB VRAM are unchanged: a new test boots `todo.gb` in libmgba, reads tile data at `0x8010` off the bus, and matches a reference expansion derived from `FONT8` alone. It passed on the unmodified base first, so it is a real oracle.
- GB parity does not cover font pixels either: two single-bit runtime mutations (plane-1 flip, dropped style-1 complement) leave parity 6/6 green while failing 2 of 3 new `gb-font` tests.
- Generated C is byte-identical for GBA, NES, ESP32 and Playdate (exactly one changed line in the GB output); `todo.gba` and `todo.nes` keep their baseline hashes.
- `bun test vapor/tests/`: 76 pass / 0 fail (was 71), including four-way parity.

## Provenance

Autonomous fleet run: Scout-measured candidate (report on the fork branch `lfkdsk/pocketjs:fleet/scout-artifact-size`), Builder implementation, one cross-family review that independently rebuilt and re-ran the mutations.

### DIFF
--- a/vapor/compiler/compile.ts
+++ b/vapor/compiler/compile.ts
@@ -2386,7 +2386,7 @@ class AppCompiler {
       [...this.strLits.keys()].reduce((a, s) => a + s.length + 1, 0) + this.title.length + 1;
     const pairCount = this.styleTable.pairs.length;
     const fontBytes =
-      this.target.name === "esp32" || this.target.name === "playdate" ? 95 * 8 : 95 * 32;
+      this.target.name === "gba" || this.target.name === "nes" ? 95 * 32 : 95 * 8;
     const styleBytes =
       this.target.name === "gba"
         ? pairCount * 16 * 2 + pairCount + 3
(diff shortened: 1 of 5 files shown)


lfkdsk:

PR #394, opened 2026-09-08
perf(vapor): compile the three GB sdcc translation units in parallel

lfkdsk:
## What

`buildGbRom` compiled its three sdcc translation units serially (`vapor/compiler/rom.ts`). The compiler stage is the bulk of a GB build — the Pocket Vapor compiler itself is ~6% of wall clock, `typescript.js` load ~43%, external toolchains ~48% — so this runs the three units concurrently and keeps the link order unchanged.

## Measured

| | before (median) | after (median) | delta |
|---|---|---|---|
| full `--target gb` build | 3108 ms | 1847 ms | −40.6% |
| three-unit sdcc stage alone | 2721 ms | 1592 ms | −41.5% |

Sample ranges do not overlap (serial min 3073 ms > parallel max 1882 ms). The stage is now bounded by the slowest unit (`gen_app`, ~1.6 s). On a single core the gain disappears (+1%), as expected.

## Correctness

- All three `.rel` files, `crt0.rel`, the intermediate `app.ihx` and the final `todo.gb` (32768 B) are byte-identical to the serial build across 5 consecutive parallel builds.
- Parallelizing needed two failure-handling fixes beyond the scheduling change, because `Promise.allSettled` does not abort on the first failure the way the serial `await` loop did: failures are collected and re-thrown naming the unit, and the second commit (from the cross-family review) additionally removes partial `.rel` outputs and the stale final ROM on failure and preserves stdout-only diagnostics in link order.
- New `vapor/tests/gb-build.test.ts` with an sdcc shim (per-unit delay, partial-output and stdout-only failure injection); seven guard-removal mutations each fail a distinct case. `bun test vapor/tests/`: 77 pass / 0 fail.

## Known limitation (pre-existing, not introduced here)

Two different apps building `--target gb` into the same output directory at the same time race on the shared `gen-gb` intermediates; this was observable before this change and is left for a follow-up.

## Provenance

Autonomous fleet run: Scout-measured candidate (report on the fork branch `lfkdsk/pocketjs:fleet/scout-compile-time`), Builder implementation, cross-family review that hardened failure handling on top.

### DIFF
--- a/vapor/compiler/rom.ts
+++ b/vapor/compiler/rom.ts
@@ -148,6 +148,8 @@ export async function buildGbRom(app: CompiledApp, outRom: string): Promise<{ ro
   const outDir = dirname(outRom);
   const genDir = join(outDir, "gen-gb");
   await $`mkdir -p ${genDir}`.quiet();
+  // A failed rebuild must not leave the previous ROM looking current.
+  await rm(outRom, { force: true });
   const genC = join(genDir, "gen_app.c");
   await Bun.write(genC, app.c);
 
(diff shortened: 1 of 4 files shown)


lfkdsk:

PR #397, opened 2026-09-10
feat(ipodtouch4): add persistent Pocket Runtime hot updates

HalfSweet:
The iPod touch 4 host currently embeds each guest into its IPA. Add a separate `PocketRuntime.app` that accepts `.pocket` updates over the shared 3DS Runtime protocol, keeping native installation separate from guest iteration.

- Add `bun ipodtouch4:runtime` commands for native build/deployment, USB pairing, guest packaging, push, source watching, DevTools, status and the device acceptance pass. USB uses the existing pinned SSH forwarder; LAN supports paired discovery and explicit addresses.
- Validate packages before replacing the guest, wait for its first GLES presentation, and commit active/last-good generations. Release the old QuickJS realm, GPU resources and touch identities between guests. Recover from guest failures and damaged stored packages, with embedded Clear as the final recovery guest.
- Keep the native listener available when a guest fails, bound QuickJS execution and socket work, and reconnect across foreground/background transitions.
- Add real QuickJS/C ABI integration tests and run them in the Linux/macOS native harness workflow. Document the portrait viewport, reload semantics and deployment commands.

## Review follow-up

1. **One PKRT server for every host.** `engine/runtime/dev_server.c` now owns pairing, the hello/ack handshake, frame dispatch, control records, uploads, screenshot streaming, idle timeouts and counters. `hosts/3ds/src/devserver.c` keeps only libctru sockets, `osGetTime` and linear screenshot buffers; `engine/runtime/dev_wire_posix.c` is the BSD socket pump for UIKit shells and the host harness. Target id, label and port are configuration (the label comes from the shell: `POCKET_DEV_RUNTIME_LABEL`). The semantics that differed are unified on both hosts: a wrong key receives its status-2 ack before the close, unknown frame types are skipped, a committed upload survives a disconnect until the frame loop takes it, the PONG is reserved until the queue drains, and the idle timeout is 3 s before the hello and 15 s after.
2. **Plan admission out of `quickjs-c`.** `pocket_runtime_validate_plan` is gone. `engine/core/src/plan.rs` validates a plan section against a `PocketTargetContract` exposed through `engine/ui-cabi` as `pocket_package_validate_plan`; `tools/target-contract.ts` renders that contract into `pocket_target_contract.h` for each native build from the target registry (`IPODTOUCH4_DEV_CONTRACTS`) and the verified plan. The `validate_plan` host callback stays; the shell and the test fixture both call the package layer with the generated contract.
3. **Generic client split.** `tools/pocket-runtime-client.ts` carries only PKRT client/session code and emits raw screenshot surfaces; PICA200 decoding and the dual-screen PNG live in `tools/3ds-runtime-client.ts` (`render3dsScreenshotPng`). The PNG encoder moved to `tools/png.ts`; `tests/png.ts` re-exports it and the other tools now import from `tools/`.
4. **Device acceptance (done on hardware).** iPod4,1, iOS 6.1.6 (10B500), USB SSH tunnel, build `e5d48dd4…`, receipt `.pocket-build/validation/ipodtouch4-runtime/2026-09-12T09-26-48-424Z/` (ignored directory; numbers below). `bun ipodtouch4:runtime acceptance --app clear --auto-background` runs the pass and writes the receipt.

   | step | result |
   | --- | --- |
   | replacement ×5 (Clear, distinct hashes) | all `accepted`; generation 19 → 24; boot 2012–2071 ms, first frame 2379–2437 ms |
   | first-frame failure | `rejected`; `active` unchanged |
   | later-frame failure (throws after 90 frames) | `accepted`, then previous package `active` again after 3999 ms; failed hash absent from `lastGood` |
   | reconnect | `getTree` and `eval ui.__host` answer on the new connection |
   | background / foreground (SpringBoard app switch, 10 s) | disconnect and reconnect observed; package still `active`; push after return `accepted` |
   | resident memory (`resident_kb`, `task_info`) | 22168 KiB before, 22256 KiB after seven swaps in one process |

   Two findings from the device, both in this branch now:
   - The 2 s guest boot budget rejected every guest on the iPod (`InternalError: interrupted`): Clear needs 2.0–2.8 s to evaluate and mount there. Budgets are build constants (`POCKET_DEV_GUEST_BOOT_BUDGET_MS` 15 s, `POCKET_DEV_GUEST_TURN_BUDGET_MS` 3 s; the harness compiles 2 s / 500 ms), and the receipts/status report measured `bootMs`/`firstFrameMs`.
   - iOS 6 terminates this shell in the background rather than suspending it (`applicationWillTerminate:` arrives after the background transition; the pre-existing single-app shell records the same `terminated` state). The relaunch reloads the committed generation and reopens the listener, so the round trip works, but guest memory state does not survive it and a return costs one boot (2.4 s). Documented in `docs/IPODTOUCH4.md`; the resign-active path no longer issues GL calls after entering the background. Why UIKit chooses termination for this binary is a separate question about the legacy shell.

Validation:
- `tests/pocket-runtime-server.test.ts` (new): the shared server replays one transcript without a transport, then the desktop client drives the 3DS libctru pump (compiled on the host with `tests/fixtures/3ds-stubs/3ds.h` and the console's literal `sdmc:` paths) and the POSIX pump through one scenario: discovery, rejected key, heartbeats, control echo, unknown frame, upload accepted, bad offset, rejected footer, dual-screen screenshot (3DS), malformed control closes.
- `tests/ipodtouch4-runtime.test.ts` with the real QuickJS: pairing, fragmentation, interrupted transfers, malformed/foreign packages, wrong presentation / unknown feature / host extension rejected by the target contract, eval/first-frame/later-frame failures, execution timeouts, restart recovery, compiled Clear and DevTools queries. It skips when the pinned sources or nightly are absent and is mandatory whenever `POCKETJS_QUICKJS_SOURCE` is set (the harness workflow).
- `cargo test` for `engine/core` (`plan::` tests) and `engine/ui-cabi --features harness-access`; the Nintendo 3DS host builds with devkitARM in the `devkitpro/devkitarm` container with no warnings from `dev_server.c`/`devserver.c`; 3DS wire/state/profile tests, `tests/test-suite.test.ts`, `bun tests/contract.ts`, `git diff --check` and `tsc --noEmit` (only pre-existing errors in unrelated tests) pass.
- Branch rebased onto current `main`.


doodlewind (commented):
Before merging, I’d like to tighten a few architecture boundaries. The implementation direction is good, but this PR is large enough that I don’t want the iPod runtime to become a second runtime architecture that drifts from 3DS.

My merge gate would be:

1. **Unify PKRT server semantics.**
   We now have two independent PKRT server state machines: `hosts/3ds/src/devserver.c` and `engine/runtime/dev_server.c`. They already differ in protocol behavior (for example unknown frames / auth rejection behavior). Please move the common auth, framing, control, upload, timeout, and dispatch semantics into one shared implementation, leaving only platform/socket-specific integration in the hosts. Target name / discovery label should also be configuration, not hard-coded iPod-specific values in the generic runtime layer.

2. **Move target/build-plan admission policy out of `quickjs-c`.**
   `pocket_runtime_validate_plan()` currently hard-codes target policy such as `presentation: "native"` and the allowed feature set (`input.touch`, `text.glyphs.baked`). The QuickJS executor should not own device capability policy. Please keep the `validate_plan` host callback, but implement the actual validation in the package/target-contract layer or generated host-side contract code, so there is one source of truth for target capabilities.

3. **Finish the generic runtime-client split.**
   `tools/pocket-runtime-client.ts` should contain only generic PKRT client/session functionality. PICA RGB decoding, dual-screen composition, and PNG generation are 3DS-specific and should live in a 3DS-specific module. Also avoid production tooling importing `tests/png.ts`.

4. **Do one real-device acceptance pass before merge.**
   The host harness is strong, but it does not cover UIKit/EAGL lifecycle, actual presentation, installation, or USB forwarding. Please verify repeated guest replacement and recovery on a real iPod touch 4, including first-frame failure, later-frame failure, foreground/background transitions, reconnect, and several consecutive reloads. In particular, confirm the GL teardown/recreation path does not leak or retain stale resources.

I don’t think this needs a rewrite. The guest replacement/recovery manager and `.pocket` admission work are solid; the main issue is making sure the new persistent runtime becomes shared infrastructure rather than an iPod-specific fork of the existing 3DS runtime behavior.

HalfSweet:
Thanks for the review. Pushed `94a8838` (rebased onto `main`) addressing the merge gates:

1. **Unified PKRT server.** `engine/runtime/dev_server.c` is now the single state machine for auth, framing, control, upload, screenshot, timeout and dispatch semantics; `hosts/3ds/src/devserver.c` and the new `engine/runtime/dev_wire_posix.c` only pump sockets and supply the clock/buffers. Target id, label and port are configuration (the label is a shell define). Where the two implementations differed, both hosts now do the same thing: rejected hellos are acked (status 2) before the close, unknown frames are skipped, a committed upload survives a disconnect, the PONG is reserved until the queue drains, timeouts are 3 s pre-hello / 15 s idle. `tests/pocket-runtime-server.test.ts` runs the real 3DS pump on the host (libctru stubbed) and the POSIX pump through one desktop-client scenario, plus a transport-free transcript of the core.
2. **Admission policy out of `quickjs-c`.** `pocket_runtime_validate_plan` is removed. The package layer validates the plan against a `PocketTargetContract` (`engine/core/src/plan.rs` → `pocket_package_validate_plan` in `engine/ui-cabi`), and `tools/target-contract.ts` generates that contract per build from the target registry + verified plan, so the desktop resolver and the device share one source of truth. The `validate_plan` host callback stays.
3. **Client split.** `tools/pocket-runtime-client.ts` is generic (raw screenshot surfaces only); PICA decoding and the dual-screen PNG moved to `tools/3ds-runtime-client.ts`. The PNG encoder now lives in `tools/png.ts`; production tools no longer import from `tests/`.
4. **Real-device pass.** Not done yet: this machine has no libimobiledevice/iproxy or the iPod's pinned SSH credentials. The exact pass (repeated replacement, first-frame/later-frame failure, foreground/background, reconnect, GL teardown with a resident-size check) is now written down in `docs/IPODTOUCH4.md` § Device acceptance so it can be run and recorded here before merge.

Also verified the 3DS host still builds with devkitARM (`bun tools/3ds.ts 3ds-demo`) with no warnings from the shared files.

HalfSweet:
Device pass done on an iPod touch 4 (iOS 6.1.6) over USB, pushed as `779beab`. The PR description carries the receipt table; the short version:

- Repeated replacement ×5, first-frame failure, later-frame failure with recovery (4.0 s), reconnect, a background/foreground round trip and resident memory (22168 → 22256 KiB after seven swaps in one process) all pass. `bun ipodtouch4:runtime acceptance --app clear --auto-background` reproduces the pass and writes a receipt under `.pocket-build/validation/`.
- Two things the device taught us, both fixed/documented in the branch: the 2 s boot budget rejected every guest (Clear needs 2.0–2.8 s to boot on the A8; budgets are build constants now, with measured `bootMs`/`firstFrameMs` in the receipts), and iOS 6 terminates this shell in the background instead of suspending it (`applicationWillTerminate:`; the pre-existing single-app shell behaves the same). The relaunch reloads the committed generation and the listener returns, so the round trip works, at the cost of one boot per return. The resign-active path no longer issues GL calls after entering the background. Why UIKit picks termination for this binary looks like a legacy-shell question rather than a Runtime one; happy to dig into it separately.

PR #407, opened 2026-09-11
feat(d211-linux): add ArtInChip D211 fbdev host and cross-build

boa-z:
Adds a **private `d211-linux-dev` target** and native host for the **ArtInChip D211DBV** running Luban Linux 5.10 on RISC-V. Software rendering only — no EGL, GLES, DRM, or Wayland.

PocketJS hero on the full 800×480 panel, Count: 34 after thirty-three taps

## What landed

- `hosts/d211-linux/` — fbdev presentation with `FBIOGET_VSCREENINFO`/`FBIOGET_FSCREENINFO` probing and a runtime channel-bitfield check; `FBIOPAN_DISPLAY(yoffset=0)` at startup so the host owns the scanned-out page; capability-probed evdev touch (`EVIOCGNAME`/`EVIOCGBIT`/`EVIOCGABS`, MT axes preferred); bounds hit resolved once at the down edge; damage-rect presentation scaled by raster density.
- `tools/d211-linux-profile.ts` — host ABI 11, `platform: "linux"`, takeover form, **logical viewport equal to the 800×480 panel at raster density 1**, capabilities `input.touch` and `text.glyphs.baked`. Stays out of `POCKET_TARGETS` until the hardware acceptance receipt is complete.
- `tools/d211-linux.ts` + `tools/d211-linux/build-runtime.sh` — builds on the Ubuntu Luban builder with the Xuantie GCC wrapper and sysroot; **final link through LLVM LLD from `nightly-2026-07-02`** because binutils 2.35 rejects modern RISC-V ELF attributes; JSON receipt records the commit, toolchain versions, sysroot, and digests.
- `tools/d211-linux/remote-build.sh` — macOS → builder sync, build, and artifact fetch in one command.
- `apps/d211-demo/` — Hero wrapper on the wide-surface (`largeLayout`) variant.
- `fix(ui-cabi)` — declares the exported scaled raster entry points (`ui_render*_scaled`) in `pocket_ui_cabi.h`.
- `test(d211-linux)` — profile, host-probe, and build-bridge tests plus the demo admission-matrix entry.

## Hardware validation

D211DBV, Luban Linux 5.10.44, build `472f7e072276bd38`:

| Measurement | Value |
| --- | --- |
| Boot (QuickJS init + `app.js` eval) | 353 ms |
| Frame rate | 59.9 fps |
| Damage | 4,080 attempts, 0 failures, 2 full redraws |
| tick / render / present | avg 1.34 ms / 2.17 ms / 0.04 ms |
| Host RSS / `MemAvailable` | 6.8 MB / ~17.8 MB, stable through touch input |
| Touch | 33 clean down/up pairs at 800×480, one action per tap (`Count: 34`) |

The captured 800×480 frame has no black pixels and the channel order matches the guest BGRA output. `/tmp` is a tmpfs on this board; an install under `/tmp` OOM-killed the host at 11 MB available, so `deploy` installs to `/opt/pocketjs` on the UBIFS rootfs, where the bundle stays reclaimable as page cache.

One correctness fix belongs to the host build: `main.c` was compiled with its fallback raster density while `pocket_runtime.c` received the plan value, so a density-1 plan rendered a scaled(2) surface. The define now reaches both translation units.

## Reproduce

```sh
# Ubuntu builder with the Luban SDK built for d211:
bun tools/d211-linux.ts setup
bun tools/d211-linux.ts build
bun tools/d211-linux.ts doctor

# macOS with the D211 on ADB:
tools/d211-linux/remote-build.sh
bun tools/d211-linux.ts deploy
bun tools/d211-linux.ts stop-ui
bun tools/d211-linux.ts run
```

## Build identity

| Input | Pin |
| --- | --- |
| Rust | `nightly-2026-07-02`, target `riscv64gc-unknown-linux-gnu` (rustc 1.98.0-nightly) |
| GCC | Xuantie-900 V2.10.1, GCC 10.4.0 (Luban wrapper + sysroot) |
| Linker | LLD 22.1.8 from the same nightly (`gcc-ld/ld.lld`) |
| QuickJS | `ba5bdd0dc013518768e76cd9e05cd30ed53dd35b` (2026-06-04) |
| Executable | `ec5ab5cadc5a87762188ac9be0128a17eeb7be5c0a3e02e9f664d1cd088d919a` (2,818,360 bytes) |
| Core library | `5df82ab904313f300ff5f5d8f05f3b11f9a501fc8b916ac1183e3544b1dabdda` |

## Out of scope

Camera (TP9950), audio, CAN, and ADC are not touched. The `test_lvgl` autostart is still stopped manually before launch. `d211-linux-dev` promotes to a registered target after the remaining receipt items (multi-contact input, launch integration).

Baseline is current `main`; the handover plan's `v0.10.1` suggestion predates `v0.11.0`.


doodlewind:
Would be better to keep this in your fork, we are not yet prepared to support all devices in the mainline.

boa-z:
> Would be better to keep this in your fork, we are not yet prepared to support all devices in the mainline.

okay

PR #414, opened 2026-09-12
fix(solid): install scheduler globals in the solid entry

### LINKED ISSUE #413: Solid entry does not install scheduler globals (setTimeout/clearTimeout/console) on timer-less hosts
## Layer

Guest framework runtime entry for the Solid framework: `framework/src/index.ts`.

Not the host and not the compiler: QuickJS hosts (e.g. the `d211-linux` target) only provide the frame callback / input / draw ops, and deliberately expose no JS timers. `framework/src/scheduler-polyfill.ts` exists exactly for that, but only the Vue Vapor and Octane entries import it — the Solid entry defines `queueMicrotask` inline and stops there.

## Symptom

On a QuickJS host, any Promise-based code that touches `setTimeout` before the first frame throws:

```
ReferenceError: setTimeout is not defined
```

`mount()` still renders, so the app looks alive. What breaks is everything that awaits:

- an IPC/transport handshake that arms a timeout (`new Promise(... setTimeout ...)`) rejects immediately;
- the rejection is typically swallowed at mount (`void connect()`), leaving the app in a "not connected" state;
- later user callbacks that depend on the connection throw, and the host records them as frame failures.

## Reproduction

1. Build any Solid app for a QuickJS target;
2. call `setTimeout(() => {}, 1)` during `onMount` (or use the documented `.text()`-style promise pattern around an adapter that arms a timeout);
3. run on the target: ReferenceError before the first frame.

Observed on an ArtInChip D211 (RISC-V, QuickJS) with a Solid app whose platform adapter arms a handshake timeout: `platform.connect()` rejected, and pressing a navigation button threw inside the input callback, canceling the rest of the callback.

## Fix

Import the shared polyfill at entry evaluation time in `framework/src/index.ts`:

```ts
import "./scheduler-polyfill.ts";
```

and drop the now-redundant inline `queueMicrotask` shim (the polyfill installs `queueMicrotask`, `setTimeout`, `clearTimeout`, and `console`). PR follows.

boa-z:
Fixes #413.

## What

- import the shared `./scheduler-polyfill.ts` at evaluation time in `framework/src/index.ts`;
- drop the inline `queueMicrotask` shim the polyfill supersedes (it also installs `setTimeout`, `clearTimeout`, and `console`).

## Why

QuickJS hosts expose no timers; the Vapor/Octane entries already import this polyfill, the Solid entry only defined `queueMicrotask`. Promise-based code that arms a timeout before the first frame (IPC handshake, retry) throws `ReferenceError: setTimeout is not defined`.

## Validation

- On an ArtInChip D211 (QuickJS, `d211-linux` abi 11): Solid app's transport handshake rejected before this change (and a later button callback threw inside input handling); with the patch the handshake resolves and presses dispatch normally. Same app, same host build.
- Diff is one import plus the removal of the redundant inline shim; no behavioral change for hosts that already provide the globals (guards in the polyfill).

### DIFF
--- a/framework/src/index.ts
+++ b/framework/src/index.ts
@@ -6,15 +6,11 @@
 // renderer's end-of-frame sweep [R] — so Solid effects triggered by input run
 // before detached subtrees are destroyed.
 
-// queueMicrotask polyfill (QuickJS lacks it; Solid's resource/transition paths
-// reference it lazily, so installing at module-eval time is early enough).
-if (typeof (globalThis as { queueMicrotask?: unknown }).queueMicrotask !== "function") {
-  (globalThis as { queueMicrotask?: (fn: () => void) => void }).queueMicrotask = (
-    fn: () => void,
-  ) => {
-    Promise.resolve().then(fn);
-  };
-}
+// Runtime scheduler globals (queueMicrotask/setTimeout/clearTimeout/console).
+// QuickJS provides none of them and the Solid entry needs the same shims the
+// Vapor/Octane preludes install: Promise-based user code (IPC handshakes,
+// timeouts) calls setTimeout during mount, before any frame runs.
+import "./scheduler-polyfill.ts";
 
 import {
   detectHost,


doodlewind:
This doesn't match how PocketJS works.

PR #416, opened 2026-09-12
feat(quickjs-c): host audio and backlight module ops

boa-z:
## What

The C runtime (`engine/quickjs-c`) could install only the `ui` surface, so native hosts had no way to expose device services to the guest. Two optional host op tables, filled before boot:

- **PocketAudioOps → `globalThis.audio`** with the method set the audio spec already pins (`contracts/spec/audio.ts`: createStream/destroyStream/writePcm/play/pause/stop/setVolume/endStream/poll). Hosts without the table leave the global absent and the framework player stays the silent no-op it is today, so existing targets are byte-identical.
- **PocketBacklightOps → `globalThis.backlight` {`get()`, `set(percent)`}** — pinned here until a display spec exists.

Both follow the existing `install_host` shape. `writePcm` borrows the guest buffer for the synchronous call, matching the spec's ownership rule; event lines are JSON per the spec.

## Why

The d211-linux host (PR #407) needs real UI sounds and panel brightness for a product app; without a host-module hook the only alternatives were product logic in the runtime or a second JS engine binding. The same hook unblocks any C host with device audio.

## Validation

- Build with the Luban toolchain passes (with the d211 host implementation stacked on #407: `feat/d211-host-modules`).
- On the device (ArtInChip D211, ALSA + sysfs backlight): the guest loads a WAV from the pak, then `createStream(22050,2)` → `writePcm` → `play` (aplay child) → `credit` event → `endStream`; brightness 0-100 maps onto `/sys/class/backlight` and the panel changes.
- No behavior change for hosts that pass no tables.

### DIFF
--- a/engine/quickjs-c/pocket_runtime.c
+++ b/engine/quickjs-c/pocket_runtime.c
@@ -94,6 +94,8 @@ static char reported_action_name[POCKETJS_ACTION_NAME_CAPACITY];
 static int32_t reported_action_value;
 static unsigned long reported_action_sequence;
 static int runtime_failed;
+static const PocketAudioOps *audio_ops;
+static const PocketBacklightOps *backlight_ops;
 #ifdef POCKET_SVC_WIRE
 /* spec SVC_POLL_BUF (8192) + terminator: one svcPoll batch. */
 static char svc_poll_buffer[8193];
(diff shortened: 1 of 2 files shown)


doodlewind:
See #407

PR #417, opened 2026-09-12
feat(d211-linux): audio and backlight host modules

boa-z:
**Stacked / draft: do not review before #407 (d211 host) and #416 (quickjs-c host-module hook) merge.** This branch carries both until then; after they land it rebases to the single audio/backlight commit.

## What

- `hosts/d211-linux/audio.c|h`: `PocketAudioOps` on one `aplay -t raw` child per stream plus a feeder thread over a 16384-frame ring — credits follow consumption, starvation is edge-reported as `underrun`, `endStream` drains then reports `ended`, volume is a soft gain.
- `hosts/d211-linux/backlight.c|h`: `PocketBacklightOps` over `/sys/class/backlight/<name>/brightness`, percent 0-100.
- `main.c` installs both tables before boot; `build-runtime.sh` compiles the new sources and links `-lpthread`.

## Validation (device)

- Boot: `host modules: audio + backlight`, `audio createStream handle=0 rate=22050 ch=2` (framework player loaded `audio:wav.btn` from the pak).
- Button press: `writePcm` → `play` (aplay child) → poll `{"t":"credit",...}` → `endStream` when the 4984-frame clip drains.
- Brightness: UI slider writes sysfs (6 → 1 for a ~20% drag); panel dims.


boa-z:
Validation update (d211, ALSA): the credit queue originally reported one event per fed chunk, which flooded the 32-slot queue and dropped the `ended` event — the guest then stalled after one ring. Fixed with sequence-based coalescing (occupancy changes counted, not compared), a feeder paced by the stream rate (device clock) and a 150 ms aplay buffer so `ended` aligns with actual playback end. With that, the product's alarm voices loop exactly at clip length (overspeed 3.2 s vs 3.19 s clip, reverse 1.8 s vs 1.82 s) and the startup/button sounds play. Also switched the host's asset loading to mmap (file-backed pages are reclaimable; the 8.25 MB pak now leaves ~15 MB available on the 54 MB board instead of OOM-killing at ~8 MB).

doodlewind:
See #407

PR #422, opened 2026-09-13
feat(psp): expose allocator telemetry and arena diagnostics

psplumberdev:
PSP applications need to distinguish QuickJS request pressure from reusable arena capacity. This PR adds process-lifetime QuickJS allocation counters and a non-allocating arena free-space snapshot, without introducing product logic into the host.

Follow-up integration:
- Preserve the last failed nonzero request across malloc(0), and saturate the call counter on long-running devices.
- Document requested bytes versus allocator overhead, owning-thread access, and the O(free-block-count) cost of the diagnostic snapshot.
- Exercise production arena/qjs allocator code through host substitutes in debug and optimized Rust. Cover in-place/moving/shrinking realloc, failed realloc with preserved data, zero-size calls, free-list splitting, permanent storage, and diagnostic accounting.
- Add `hosts/psp/examples/allocator-check.rs`: an interactive PSP test of an 18,200,001-byte permanent buffer and real QuickJS allocation/GC/runtime teardown. CROSS repeats the workload without allocating another permanent buffer; HOME exits.

This branch includes #423 for the combined tests and device example. Merge #423 first; its changes will then disappear from this PR's main-relative diff. Original contributor commits are retained, with integration/follow-up commits appended; use merge commits rather than squash/rebase.

Validation:
- 22 allocator cases passed (11 scenarios in debug and optimized Rust), plus 9 test-suite registration checks: 31 passed, 0 failed.
- TypeScript check and patch whitespace check passed.
- Pinned PSP release builds of both Hero and allocator-check passed. Generated binaries remain outside Git.
- Linux/macOS PSP allocator and native harness checks, plus dynamic-text checks, run on the PR.
- Physical PSP acceptance passed: the maintainer confirmed the combined allocator example after PSPLINK deployment. The captured screen showed PASS round 1, live QuickJS bytes returning to zero, and READY with no FAIL; user interaction acceptance was then confirmed.

Device build: #422 `174b5c5d02ee3de733a5962884e990a0831ab68f`, #423 `dc4fce0ea95d0634c4fa4f014081de79fd106e53`.
Command: `POCKETJS_ALLOCATOR_REVISION='422:174b5c5d 423:dc4fce0e' bun tools/psp.ts hero --release --example allocator-check`.
Allocator EBOOT SHA-256: `0fd19fa2ffe8b5488fbca2a2169d5f5fbde0aba3ade9bbc3bd478ae207a21431`.
[...]

PR #423, opened 2026-09-13
fix(psp): avoid size-class rounding for permanent map buffers

psplumberdev:
Large process-lifetime buffers should not consume a power-of-two size class. `arena::alloc_permanent(size, align)` reserves the requested bytes plus alignment padding from the existing arena tail, allowing an approximately 18.2 MB map buffer to fit without reserving a 32 MB recyclable block.

The API rejects zero size, invalid alignment, overflow and insufficient tail space without advancing the bump pointer. Its safety contract now states owning-thread access, initialization requirements, process lifetime, and the prohibition on ordinary deallocation or ownership through Box/Vec.

Validation:
- 12 host cases pass (6 scenarios in debug and optimized Rust), compiling production arena.rs with substituted PSP partition calls. Coverage includes large allocation, OOM/overflow, alignment, recycled blocks, exact-fit exhaustion and failed kernel reservation.
- Test registration guard passes; Linux/macOS checks run in the new PSP allocator workflow.
- PR #422 builds on this branch and includes a combined interactive PSP allocator acceptance example and a normal Hero build. Physical PSP acceptance passed: the maintainer confirmed the combined allocator example after PSPLINK deployment.

Original contributor commits remain intact. Merge this PR before #422, using a merge commit to retain their history. The downstream OpenStrike integration is reviewed separately.

Hardware acceptance covers #422 `174b5c5d02ee3de733a5962884e990a0831ab68f` and #423 `dc4fce0ea95d0634c4fa4f014081de79fd106e53`. Deployed PRX SHA-256: `3eaf5c10e2b97c92596a783670de8db323e0eb8608d0ac1fbded3c553296ba42`. Device model was not recorded; this does not claim a downstream game acceptance.


doodlewind:
Thanks for the contribution! This is a valuable improvement for large buffers on memory-constrained PSP hardware. We have completed hardware acceptance and merged this while preserving your original commits. I appreciate you bringing the real-world OpenStrike use case upstream. Feel free to DM @pocket_js on X — I would love to connect and hear more about your work!

PR #432, opened 2026-09-17
fix(ipodtouch4): reject failed digest checks during installation

lfkdsk:
The iPod touch 4 deployment script now accepts both `SHA256(...)` and `SHA2-256(...)` digest output and rejects a failed OpenSSL command even when it prints the expected hash. Archive failures stop before migration; installed-file verification failures retain the recovery journal for retry.

Tests use real host OpenSSL plus injected output/status faults to cover both prefixes, malformed output, and recovery after a nonzero digest command. The strict-C11 package fix is already on main and is not duplicated here.

Validation: 19 tests passed across installation, iPod touch 4 profile, and the shared TS/C/Python package corpus on macOS. Linux and macOS native harness checks run on this PR.


doodlewind:
Thanks for the contribution!

Issue #415, opened 2026-09-12
d211-linux: rendering a PSM 5650 texture kills the app (PSM 4444/8888 render)


boa-z:
## Layer

Guest pak texture format → core/host raster path on the `d211-linux` target (ArtInChip D211DBV, QuickJS host, 64 MB RAM).

## Symptom

An image entry baked with psm 0 (`PSM_5650`, RGB565) works through the pak build but the app process dies the moment the texture is drawn. The same image renders with psm 3 (8888) and psm 2 (4444).

## Reproduction

1. App declares an image and sets its format via `images.json`:
   `{ "assets/pak/frame.png": { "psm": 0 } }`;
2. Build for `d211-linux-dev`, deploy, run;
3. The process disappears right after the first frame that includes the image (no error line, no JS exception — exit via signal; the framebuffer keeps the previous frame).

Measured on the same binary and image:
- psm 3 (8888) → renders (smooth);
- psm 2 (4444) → renders, visible banding on gradients;
- psm 0 (5650) → process dies (SIGKILL/OOM or crash in the upload/draw path).

## Impact

Full-screen opaque artwork (charging animation frames, boot logo, photo backgrounds) is 800×480. With 8888 it needs two 512×512 tiles per image (2 MB texmem), which exceeds the 64 MB budget after a few frames; 5650 would halve that and is the natural format for opaque art on this target.

## Notes

- Texture dims are already pow2 ≤512 and fully opaque (the encoder's 5650 precondition is satisfied).
- Could be an unsupported psm in the D211 upload/conversion path, or an allocation failure caused by an intermediate 8888 buffer; a clear error instead of a dead process would already help.
- Happy to test a patch against the same device.

doodlewind:
See #407

PR #433, opened 2026-09-17
test(suite): enforce tape assertions and cover omitted tests

lfkdsk:
Tape replay now rejects malformed or partial assertion files before building the app. Missing values and repeated `--assert` options are errors, and a valid assertion checks every frame instead of reporting success after a partial comparison.

The test suite runs 11 previously omitted test files and the 180-frame replay, and checks that every root TypeScript test belongs to a suite stage or an explicit workflow. C ABI texture coverage exercises 5650, 4444 and 8888 uploads/draws in release and debug builds, including two 512×512 tiles. These tests run with the pinned nightly in the Linux/macOS native harness; the general release suite retains its stable Rust requirement.

Validation on macOS: all 14 suite stages passed, TypeScript passed, 73 focused assertion/suite tests passed, and both C ABI texture profiles passed. Linux/macOS native harness checks run on this PR. The six web snapshots and tape hashes already match main, so this PR does not replace those assets. The host-side texture tests do not establish a fix for the physical D211 crash reported in #415.


doodlewind:
Thanks for the contribution!

PR #436, opened 2026-09-17
perf(test): cut idf-incremental real builds from five to three

lfkdsk:
## What

`tests/idf-incremental.test.ts` ran five real ESP-IDF package builds to prove incremental behaviour; three are enough. Every existing expectation is kept (mapped one by one in the fleet report), and the file's wall clock drops accordingly.

## Verification

- `bun test tests/idf-incremental.test.ts` green; full suite 14/14 stages on the integration branch.
- Cross-model review (fleet 863) confirmed the expectation-by-expectation mapping.


HalfSweet (approved):
LGTM

PR #445, opened 2026-09-17
perf(pocket3d): reduce PSP world rendering and collision work

doodlewind:
OpenStrike's inferno view submits many off-screen triangles from PVS-visible leaves and traces every solid brush for each character sweep. These costs leave too little time to maintain game speed on PSP once simulation follows real time.

This change adds conservative bounds for each cooked face and each brush collision hull, uses stack storage for slide planes, and skips the stair search when a clear horizontal sweep already covers the requested distance. PSP world rendering builds optional triangle strips without changing vertices, textures, lightmaps, or the cooked file format. Strips preserve triangle multiplicity and require the existing two-sided world state.

The strip cache has a 512 KiB index budget and a 256 KiB range budget, with less than 64 KiB of conversion scratch. Allocation failure, oversized runs and exhausted budgets keep the original triangle list. Large draws split within the 64 KiB frame-pool allocation limit. Renderer statistics distinguish source triangles from submitted indices.

Validation:
- 19 BSP unit tests passed. Three opt-in audits passed against eight locally supplied cooked maps: 64,000 exact trace comparisons, 115,200 exact character-step comparisons across two movement configurations, and triangle-multiset checks for every cached strip. Synthetic ground/slope tests also cover floating-point lerp rounding.
- `cargo check --locked --manifest-path engine/Cargo.toml -p pocket3d-bsp --no-default-features --features libm` passed.
- OpenStrike PSP release builds and physical PSPLINK runs passed. At the original inferno spawn view, the previous build ran about 20 FPS; the optimized renderer plus OpenStrike's real-time fixed-step loop runs about 30 FPS while advancing 60 simulation ticks per second. Dust2 remains about 60 FPS. These are fixed-view comparisons, not whole-map guarantees.
- Physical moving-camera and combat probes exercise jumping, movement, firing, reloads, damage, deaths and round transitions. Manual acceptance is still pending. PSP-1000 and the unavailable de_torn asset were not tested.
- CI now runs BSP unit tests and the no_std/libm check when BSP or GU sources change. The PSP GU backend was cross-compiled and tested on hardware locally.

Reproduce the asset-dependent audits with `POCKET3D_COOKED_MAPS=/path/to/maps cargo test --locked --manifest-path engine/Cargo.toml -p pocket3d-bsp -- --include-ignored --nocapture`. Proprietary maps, binaries, screenshots and raw benchmark logs are not committed.

OpenStrike integration will pin this commit in a separate Draft PR, following pocket-stack/open-strike#19. Keep this PR in Draft until physical acceptance.


doodlewind:
Thanks for the contribution! @psplumberdev, your work helped drive these PSP improvements, and you are credited as a co-author. The changes have passed hardware acceptance and are now merged.

